PR #2, opened 2021-07-17
Added API Endpoint to Search by Combined User First, Middle, and Last Names

kjriedel:
* Added Controller API endpoint to search for users by combined first, middle, and last names.  For example:
`http://localhost:9999/user/retrieve?name=JackPool`

* Updated User filter and specification to add ability to filter by combined first, middle, and last names. The existing logic in `AbstractSpecification::applyStringFilter` is still used to process negation("!") and wildcard ("*"/"%") characters.  Therefore, the new API `name` query parameter may contain these operators.  For example:
`http://localhost:9999/user/retrieve?name=!*Wagon`

* Added validator method to validate user name, ensuring it contains only valid characters and no space.

Ad-hoc tested with Chrome 91.0 and Postman.  Junit tests are still WIP...

### DIFF
--- a/src/main/java/com/bravo/user/controller/UserController.java
+++ b/src/main/java/com/bravo/user/controller/UserController.java
@@ -43,10 +43,33 @@ public class UserController {
     return userService.create(request);
   }
 
-  @GetMapping(value = "/retrieve/{id}")
+  @GetMapping(value = "/retrieveId/{id}")
   @ResponseBody
-  public UserReadDto retrieve(@PathVariable String id){
-    return userService.retrieve(id);
+  public UserReadDto retrieveById(@PathVariable String id){
+    return userService.retrieveById(id);
+  }
+  
+  /**
+   * Retrieves a paginated list of users whose first, middle, and last names match the specified name.
+   * The name to search may contain negation ("!") or wildcard ("*" or "%") characters
+   * 
+   * @param name Name to find - should contain first, middle, and last name without space
+   * @param page Page number to retrieve
+   * @param size Size of the page to retrieve
+   * @param response {@linkplain HttpServletResponse}
+   * @return List of {@linkplain UserReadDto}s with combined first, middle, and last names that match the search criteria
+   */
+  @GetMapping(value = "/retrieveName/{name}")
+  @ResponseBody
+  public List<UserReadDto> retrieveByName(
+	  final @PathVariable String name,
+	  final @RequestParam(required = false) Integer page,
+	  final @RequestParam(required = false) Integer size,
+	  final HttpServletResponse response) {
+	  
+	  userValidator.validateName(name);
+	  final PageRequest pageRequest = PageUtil.createPageRequest(page, size);
+	  return userService.retrieveByName(name, pageRequest, response);
   }
 
   @PostMapping(value = "/retrieve")
--- a/src/main/java/com/bravo/user/controller/UserController.java
+++ b/src/main/java/com/bravo/user/controller/UserController.java
@@ -9,6 +9,8 @@ import com.bravo.user.service.UserService;
 import com.bravo.user.utility.PageUtil;
 import com.bravo.user.validator.UserValidator;
 import java.util.List;
+import java.util.Set;
+
 import javax.servlet.http.HttpServletResponse;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.validation.BindException;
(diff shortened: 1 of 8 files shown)


src/main/java/com/bravo/user/controller/UserController.java
```diff
@@ -43,10 +43,33 @@ public UserReadDto create(final @RequestBody UserSaveDto request, final BindingR
...
+   * Retrieves a paginated list of users whose first, middle, and last names match the specified name.
+   * The name to search may contain negation ("!") or wildcard ("*" or "%") characters
+   * 
+   * @param name Name to find - should contain first, middle, and last name without space
+   * @param page Page number to retrieve
+   * @param size Size of the page to retrieve
+   * @param response {@linkplain HttpServletResponse}
+   * @return List of {@linkplain UserReadDto}s with combined first, middle, and last names that match the search criteria
+   */
+  @GetMapping(value = "/retrieveName/{name}")
```
wheelers-websites, line 62: I wouldn't expect uppercase in the URL. Perhaps something like:
* /retrieve/id/{id} and /retrieve/name/{name}

I would've preferred a common GET method that took the values as optional RequestParam's
* /retrieve?id={id}
* /retrieve?name={name}

src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
```diff
@@ -86,9 +86,10 @@ protected void applyDateTimeFilter(
         inClause.add(s.toLowerCase());
         continue;
       }
-      final String targetValue = (isNot ? s.substring(1) : s).toLowerCase();
+      String targetValue = (isNot ? s.substring(1) : s).toLowerCase();
       Predicate predicate;
       if(isLike){
+    	targetValue = targetValue.replace("*", "%"); // "like" queries must use "%"
```
wheelers-websites, line 92: Good catch! I didn't realize this bug existed in the code.

src/main/java/com/bravo/user/service/UserService.java
```diff
@@ -39,13 +41,37 @@ public UserReadDto create(final UserSaveDto request){
...
+  /**
+   * Retrieves a paginated list of users whose combined first, middle, and last names matches the specified name.
+   * 
+   * @param name Name to find with format [firstName][middleName][lastName] and no space. May contain negation ("!") or
+   * wildcard ("*"/"%") characters.
+   * @param pageRequest {@linkplain PageRequest} containing pagination parameters
+   * @param response {@linkplain HttpServletResponse} in which to store page information
+   * @return List of {@linkplain UserReadDto}s whose combined names match the search name criteria
+   */
+  public List<UserReadDto> retrieveByName(
```
wheelers-websites, line 61: This method is unnecessary. The code almost completely duplicates `retrieve`. If the `UserFilter` was created in the `UserController` then the same method could've been used.

src/main/java/com/bravo/user/validator/UserValidator.java
```diff
@@ -14,6 +17,20 @@ public void validateId(String id){
...
   }
+  
+  /**
+   * Validates a user name to ensure it is not null, not empty, and does not contain space, numbers, or special characters.
+   * 
+   * @param name User name to validate
+   * @throws BadRequestException if the name is invalid
+   */
+  public void validateName(String name) {
+	  Pattern pattern = Pattern.compile("[^a-z*%!]", Pattern.CASE_INSENSITIVE);
```
wheelers-websites, line 28: This `Pattern` could be `private static final`

src/main/java/com/bravo/user/validator/UserValidator.java
```diff
@@ -14,6 +17,20 @@ public void validateId(String id){
...
+	  if (ValidatorUtil.isInvalid(name) || pattern.matcher(name).find()) {
```
wheelers-websites, line 29: Why did you use `matcher(name).find()` instead of `matcher(name).matches()`?

src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
kjriedel: Thanks.

src/main/java/com/bravo/user/validator/UserValidator.java
kjriedel: Good call.  Will update.

src/main/java/com/bravo/user/validator/UserValidator.java
kjriedel: I did try `matches()`.  It only matches against the full string.  `find()` will also match against substrings, which appeared to work properly when tested.  See:

https://stackoverflow.com/questions/4450045/difference-between-matches-and-find-in-java-regex

src/main/java/com/bravo/user/controller/UserController.java
kjriedel: Yep, I get that.  I wasn't sure if it was acceptable to modify the existing code to work with RequestParams instead of PathVariables or just add an additional method, so I went with the latter.  Will update though.

src/main/java/com/bravo/user/service/UserService.java
kjriedel: Yep, will update.

kjriedel:
Changes based upon review comments have been pushed.

PR #3, opened 2021-08-03
Implemented all items on checklist

marksguitars:


wheelers-websites:
I want to start my saying that we are not expecting any additional commits. Feel free to make them if you wish, but that's not an expectation that we have. With that said I will go through your changes and leave technical comments on them.

src/main/java/com/bravo/user/dao/model/User.java
```diff
@@ -18,6 +18,26 @@
   @Column(name = "id")
   private String id;
 
+  public void setFirstName(String firstName) {
```
wheelers-websites, line 21: These getters/setters are not necessary since the [Lombok](https://projectlombok.org/) `@Data` annotation provides them.

src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
```diff
@@ -68,6 +68,17 @@ protected void applyDateTimeFilter(
     predicates.add(path.in(values));
   }
 
+  protected <X, Y  extends Collection<X>> void applyInLengthFilter(
```
wheelers-websites, line 71: 1. This `AbstractSpecification` is only used for searching entities so validation at this layer is not practical. 
2. Although nice use of Generics +1.

src/main/java/com/bravo/user/dao/specification/UserSpecification.java
```diff
@@ -4,7 +4,9 @@
 import com.bravo.user.model.filter.UserFilter;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Path;
 import javax.persistence.criteria.Root;
+import java.util.Set;
 
 public class UserSpecification extends AbstractSpecification<User> {
```
wheelers-websites, line 11: Similar to `AbstractSpecification` comment. This class is only used for searching and invoked from the service layer. The validation should happen before the code reaches this point.

src/main/java/com/bravo/user/model/dto/UserReadDto.java
```diff
@@ -8,5 +8,19 @@
 
   private String id;
   private String name;
+  private Integer phoneNumber;
   private LocalDateTime updated;
+
+  public void setName(String name) {
```
wheelers-websites, line 14: These getters/setters are not necessary since the [Lombok](https://projectlombok.org/) `@Data` annotation provides them.

src/main/java/com/bravo/user/dao/model/User.java
```diff
@@ -27,9 +47,37 @@
...
+  public Integer getPhoneNumber() {
+    return phoneNumber;
+  }
+
+  public LocalDateTime getUpdated() {
+    return updated;
+  }
+
+  @Column(name = "phone_number", nullable = false)
+  private Integer phoneNumber;
```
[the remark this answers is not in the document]
ryanandrewmcpherson: You wouldn't want to be able to do math on a phone number.... so why is it an integer? Also you might want to allow for characters like -,+, and () in the future .... maybe there might be a requirement to parse area codes or something like that....

src/main/java/com/bravo/user/model/dto/UserSaveDto.java
```diff
@@ -8,4 +8,23 @@
   private String firstName;
   private String middleName;
   private String lastName;
+  private Integer phoneNumber;
+
+  public String getFirstName() {
```
wheelers-websites, line 13: These getters/setters are not necessary since the [Lombok](https://projectlombok.org/) `@Data` annotation provides them.

src/main/java/com/bravo/user/model/filter/UserFilter.java
```diff
@@ -7,9 +7,40 @@
...
+    return ids;
+  }
+
   private Set<String> ids;
   private Set<String> firstNames;
   private Set<String> lastNames;
   private Set<String> middleNames;
+  private Set<Integer> phoneNumbers;
+
+  public Set<String> getFirstNames() {
```
wheelers-websites, line 20: These getters/setters are not necessary since the [Lombok](https://projectlombok.org/) `@Data` annotation provides them.

src/main/java/com/bravo/user/model/filter/UserFilter.java
```diff
@@ -7,9 +7,40 @@
...
+
+  public Integer getPhoneNumberLength() {
+    return phoneNumberLength;
+  }
+
+  public DateFilter<LocalDateTime> getDateFilter() {
+    return dateFilter;
+  }
+
+  private final Integer phoneNumberLength = 10;
```
wheelers-websites, line 44: 1. This value doesn't belong here. It should be declared in a constants or validator file. 
2. This variable should be `static` so it can be referenced without an instance of the object.

src/main/java/com/bravo/user/utility/ValidatorUtil.java
```diff
@@ -62,4 +77,16 @@ private static boolean isDateFilterValid(final DateFilter<?> date){
   private static boolean isStringValid(final String string){
     return string != null && !string.trim().isEmpty();
   }
+
+  private static boolean isIntegerValid(final Integer integer) {
+    if(integer == null || integer == 0)
+      return false;
+    String onlyDigits = "\\d+";
+    String inString = integer.toString();
+    return inString.matches(onlyDigits);
```
ryanandrewmcpherson, line 86: why are you checking if an integer contains only digits?

src/main/java/com/bravo/user/utility/ValidatorUtil.java
```diff
@@ -41,12 +41,27 @@ else if(value instanceof DateFilter<?>){
...
+    return !isValidLength(value, length);
+  }
+
+  public static <T> boolean isValidLength(T value, T length){
+    boolean isValid = false;
+    if(value instanceof Integer){
+       isValid = isIntegerLengthValid((Integer) value, (Integer) length);
+      }
+    return isValid;
+  }
```
wheelers-websites, line 63: I don't like the implementation of this method. It's not very generic. It would be better if it took a min and max range and then compared the value for numbers and the length for strings.


     public static <T> boolean isValidLength(T value, double min, double max){
          Double numeric;
          if(value instanceof Number){
               numeric = ((Number)value).doubleValue();          
          }
          else if(value instanceof String){
               numeric =(double)((String)value).length;
          }
          else {
               throw new IllegalArgumentException("helpful message");
          }
          return numeric > min && numeric < max;
     }

src/main/java/com/bravo/user/validator/UserValidator.java
```diff
@@ -37,5 +50,9 @@ protected void validateUpdate(Object o, Errors errors) {
     if(ValidatorUtil.isEmpty(instance, "id", "updated")){
       errors.reject("'request' modifiable field(s) are required");
     }
+
+    if(String.valueOf(instance.getPhoneNumber()).length() > 0 && ValidatorUtil.isInvalid(instance.getPhoneNumber())){
+      errors.reject("'phoneNumber' is must contain only digits");
```
wheelers-websites, line 55: Weird wording on this error message.

src/main/java/com/bravo/user/model/filter/UserFilter.java
marksguitars: I agree and I made a comment to your point. I just did not have time to either add a validator or be familiar enough w/your code to know the proper place to put this. static of course should have been added.

src/main/java/com/bravo/user/utility/ValidatorUtil.java
marksguitars: I was thinking in a purely linear way here in regards to testing only Integer values. With that in mind, I am not sure if using a range where I would have to set both a min and max  arg to the same value to test the ceiling of an Integer value would be more effective than my isValidLength function which when all told should handle checks on only Integers in less code than your re-factored example(unless your point was to emphasize additional data types being covered).

marksguitars:
> I want to start by saying that we are not expecting any additional commits. Feel free to make them if you wish, but that's not an expectation that we have. With that said I will go through your changes and leave technical comments on them.

Thanks for the review. I did make a couple of comments so hopefully they made sense.

PR #4, opened 2021-08-11
Add API and Service Method to Retrieve Users By Name

BroShow:


wheelers-websites:

src/main/java/com/bravo/user/controller/UserController.java
```diff
@@ -35,6 +35,18 @@ public UserController(UserService userService, UserValidator userValidator) {
     this.userValidator = userValidator;
   }
 
+  @GetMapping(value = "/{name}")
```
wheelers-websites, line 38: I would expect "retrieve" in the name here to follow the conventions of the app.

src/main/java/com/bravo/user/service/UserService.java
```diff
@@ -40,6 +41,24 @@ public UserReadDto create(final UserSaveDto request){
...
+    UserFilter filter = new UserFilter();
+    String[] names = name.split(" ");
+
+    filter.setFirstNames(Set.of(names[0]));
+    if(names.length == 2) {
+      filter.setLastNames(Set.of(names[1]));
+    } else if(names.length == 3) {
+      filter.setMiddleNames(Set.of(names[1]));
+      filter.setLastNames(Set.of(names[2]));
+    }
```
wheelers-websites, line 57: This is interesting logic. Although, I don't see a scenario where they pass more than 3 names here. Tracing this code it seems like any name after the first 3 would just be ignored.

src/main/java/com/bravo/user/service/UserService.java
```diff
@@ -40,6 +41,24 @@ public UserReadDto create(final UserSaveDto request){
...
+
+    return retrieve(filter, pageRequest, httpResponse);
```
wheelers-websites, line 59: Glad you're using the existing service method. However, the UserSpecification implementation is currently doing AND queries against the specific fields you provide. So this implementation would only work for searching users with the same "first", "first and last", or "first, middle, and last" names. It doesn't support searching users that have shared names in different positions. e.g. user1.first == user2.middle

src/main/java/com/bravo/user/utility/ValidatorUtil.java
```diff
@@ -60,6 +60,6 @@ private static boolean isDateFilterValid(final DateFilter<?> date){
   }
 
   private static boolean isStringValid(final String string){
-    return string != null && !string.trim().isEmpty();
+    return string != null && !string.trim().isEmpty() && string.matches("[a-zA-Z!*% ]+");
```
wheelers-websites, line 63: I like the use of regular expressions here. I don't like that this generic implementation in a shared validator was changed to conform to these validation rules when the rules for another String used in the app could be different. 

I would have preferred to see a unique method that validated the name specifically or an expansion on this generic method to accept an optional RegExp as a parameter.

src/main/java/com/bravo/user/validator/UserValidator.java
```diff
@@ -10,6 +10,12 @@
 @Component
 public class UserValidator extends CrudValidator {
 
+  public void validateName(String name){
+    if(ValidatorUtil.isInvalid(name)){
+      throw new BadRequestException("'name' is required");
```
wheelers-websites, line 15: Your validation logic doesn't match your error message here. The user could be providing a "name", but still see this error message if the name provided doesn't match the RegExp used. That information should be included in the error message to reduce debugging time and user frustration.

src/main/java/com/bravo/user/controller/UserController.java
BroShow: The retrieve path is already being used to pull records by ID. Additionally I think it is more common to have a path such as:
/user/{name} so you can expand on other attributes to fetch about that user, such as /user/{name}/group, or some variation of this pattern, which I have also seen OKTA use when integrating with their APIs.

src/main/java/com/bravo/user/controller/UserController.java
```diff
@@ -35,6 +35,18 @@ public UserController(UserService userService, UserValidator userValidator) {
...
+  @ResponseBody
+  public List<UserReadDto> retrieveUserByName(
+          @PathVariable String name,
```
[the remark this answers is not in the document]
BroShow: I don't see an issue with either approach. This is the approach I have seen used more commonly from my personal experience, but both ways would work.

src/main/java/com/bravo/user/service/UserService.java
BroShow: For this particular logic, I went with this approach with an understanding that in a normal scenario, I would have asked for a clarification from the stakeholder for the requirement. From my understanding the API would only take a single value for a name and that name could be any portion of the users full name. If you wanted to be able to provide a more specific search where first, middle, and last name could be provided individually, I would have suggested query parameters or perform a search similar to the POST method for "/retrieve".

src/main/java/com/bravo/user/controller/UserController.java
```diff
@@ -35,6 +35,18 @@ public UserController(UserService userService, UserValidator userValidator) {
...
+          final @RequestParam(required = false) Integer page,
+          final @RequestParam(required = false) Integer size,
+          final HttpServletResponse httpResponse){
+    final PageRequest pageRequest = PageUtil.createPageRequest(page, size);
+    userValidator.validateName(name);
```
tlynema-bravolt, line 46: Would like to see the validation before other logic.

src/main/java/com/bravo/user/service/UserService.java
```diff
@@ -40,6 +41,24 @@ public UserReadDto create(final UserSaveDto request){
     return resourceMapper.convertUser(user);
   }
 
+  public List<UserReadDto> retrieveByName(
+          final String name,
+          final PageRequest pageRequest,
+          final HttpServletResponse httpResponse){
+    UserFilter filter = new UserFilter();
```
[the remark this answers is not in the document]
tlynema-bravolt: Can it be at the class-level?

src/main/java/com/bravo/user/utility/ValidatorUtil.java
BroShow: I had a similar thought, but noticed that the only search String search parameters used were id, first name, middle name, and last name, which I thought could have the same validation.

src/main/java/com/bravo/user/service/UserService.java
```diff
@@ -40,6 +41,24 @@ public UserReadDto create(final UserSaveDto request){
...
+    String[] names = name.split(" ");
```
tlynema-bravolt, line 49: This is a straight-forward way to parse a name with spaces.  This forces a first name to be passed in though.  Is there a way to make this so it can search by any part of a name?

src/main/java/com/bravo/user/validator/UserValidator.java
BroShow: I agree, after I went back and added the additional validation for preventing numbers and characters, this message was no longer sufficient. Additionally I wanted to add that with my initial implementation of this method in order to follow the requirements on the trello board, I wanted to point out that since the value was a PathVariable, it could not be null or empty, because then it wouldn't map to the GET API.

src/main/java/com/bravo/user/service/UserService.java
BroShow: Yes, this could be done several ways. If the desire was to create a search for the name value in any of the fields, I would have provided and OR query implementation. The way I interpreted the requirement though, was that a single name value was provided to be able to search for a full or partial value within the name.

PR #5, opened 2021-08-19
Add Search by Name

unamanic:
It's been a while since I've used Criteria objects instead of JPQL. 

image
image

### DIFF
--- a/src/main/java/com/bravo/user/controller/UserController.java
+++ b/src/main/java/com/bravo/user/controller/UserController.java
@@ -45,8 +45,22 @@ public class UserController {
 
   @GetMapping(value = "/retrieve/{id}")
   @ResponseBody
-  public UserReadDto retrieve(@PathVariable String id){
+  public UserReadDto retrieve(@PathVariable String id) {
     return userService.retrieve(id);
+  }
+
+  @GetMapping(value = "/retrieve")
+  @ResponseBody
+  public List<UserReadDto> retrieveByName(
+      final @RequestParam String name,
+      final @RequestParam(required = false) Integer page,
+      final @RequestParam(required = false) Integer size,
+      final HttpServletResponse httpResponse
+  ) {
+    // validate name but allow control characters
+    userValidator.validateName(name.replaceAll("[!*%]", ""));
+    final PageRequest pageRequest = PageUtil.createPageRequest(page, size);
+    return userService.retrieveByName(name, pageRequest, httpResponse);
   }
 
   @PostMapping(value = "/retrieve")
@@ -56,7 +70,7 @@ public class UserController {
       final @RequestParam(required = false) Integer page,
       final @RequestParam(required = false) Integer size,
       final HttpServletResponse httpResponse
-  ){
+  ) {
     final PageRequest pageRequest = PageUtil.createPageRequest(page, size);
     return userService.retrieve(filter, pageRequest, httpResponse);
   }
@@ -75,7 +89,7 @@ public class UserController {
 
   @DeleteMapping(value = "/delete/{id}")
   @ResponseBody
-  public boolean delete(final @PathVariable String id){
+  public boolean delete(final @PathVariable String id) {
     userValidator.validateId(id);
     return userService.delete(id);
   }
--- a/src/main/java/com/bravo/user/controller/UserController.java
+++ b/src/main/java/com/bravo/user/controller/UserController.java
@@ -7,6 +7,7 @@ import com.bravo.user.model.dto.UserSaveDto;
 import com.bravo.user.model.filter.UserFilter;
 import com.bravo.user.service.UserService;
 import com.bravo.user.utility.PageUtil;
+import com.bravo.user.utility.ValidatorUtil;
 import com.bravo.user.validator.UserValidator;
 import java.util.List;
 import javax.servlet.http.HttpServletResponse;
@@ -58,7 +59,7 @@ public class UserController {
       final HttpServletResponse httpResponse
   ) {
     // validate name but allow control characters
-    userValidator.validateName(name.replaceAll("[!*%]", ""));
+    userValidator.validateName(ValidatorUtil.removeControlCharacters(name));
     final PageRequest pageRequest = PageUtil.createPageRequest(page, size);
     return userService.retrieveByName(name, pageRequest, httpResponse);
   }
--- a/src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
+++ b/src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
@@ -88,8 +88,7 @@ public abstract class AbstractSpecification<T> implements Specification<T> {
         continue;
       }
       final String targetValue = (isNot ? s.substring(1) : s).toLowerCase();
-      Predicate predicate = createPredicate(targetPath, targetValue, isLike, isNot);
-      predicates.add(predicate);
+      predicates.add(createPredicate(targetPath, targetValue, isLike, isNot));
     }
     if(!inClause.isEmpty()){
       applyInFilter(targetPath, inClause);
--- a/src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
+++ b/src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
@@ -96,6 +96,9 @@ public abstract class AbstractSpecification<T> implements Specification<T> {
   }
 
   protected void applyStringFilterToFields(Collection<Expression<String>> fields, String value) {
+    if(ValidatorUtil.isInvalid(value)){
+      return;
+    }
     final boolean isLike = isLike(value);
     final boolean isNot = isNot(value);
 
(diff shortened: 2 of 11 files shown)


src/main/java/com/bravo/user/controller/UserController.java
```diff
@@ -45,18 +45,32 @@ public UserReadDto create(final @RequestBody UserSaveDto request, final BindingR
...
+  @GetMapping(value = "/retrieve")
+  @ResponseBody
+  public List<UserReadDto> retrieveByName(
+      final @RequestParam String name,
+      final @RequestParam(required = false) Integer page,
+      final @RequestParam(required = false) Integer size,
+      final HttpServletResponse httpResponse
+  ) {
+    // validate name but allow control characters
+    userValidator.validateName(name.replaceAll("[!*%]", ""));
```
tlynema-bravolt, line 61: I like that the validation is first.  The replaceAll... should that be a method that other calls can utilize?    Also, is there a potential NPE here?

src/main/java/com/bravo/user/model/filter/UserFilter.java
```diff
@@ -2,14 +2,18 @@
...
 
 @Data
+@Builder
 public class UserFilter {
 
   private Set<String> ids;
   private Set<String> firstNames;
   private Set<String> lastNames;
   private Set<String> middleNames;
+  private String name;
```
tlynema-bravolt, line 17: Can you add the pros and cons of adding name here instead of its own class or trying to divide the name into its parts {firstName, lastName, MiddleName}.

src/main/java/com/bravo/user/service/UserService.java
```diff
@@ -48,6 +48,22 @@ public UserReadDto retrieve(final String id){
...
+      final String name,
+      final PageRequest pageRequest,
+      final HttpServletResponse httpResponse
+  ){
+    final UserSpecification specification = new UserSpecification(UserFilter.builder()
+        .name(name)
+        .build());
+    final Page<User> userPage = userRepository.findAll(specification, pageRequest);
+    final List<UserReadDto> users = resourceMapper.convertUsers(userPage.getContent());
+    LOGGER.info("found {} user(s)", users.size());
```
tlynema-bravolt, line 61: Thank you for using {} notation :+1:  .   In my view, this is a debug and not an info though.

src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
```diff
@@ -79,28 +80,57 @@ protected void applyDateTimeFilter(
...
+    final boolean isNot = isNot(value);
+
+    final String targetValue = (isNot ? value.substring(1) : value).toLowerCase();
+    Predicate[] filterPredicates = fields.stream().map(field -> createPredicate(field, targetValue, isLike, isNot)).toArray(Predicate[]::new);
+
+    if(isNot){
+      predicates.add(criteriaBuilder.and(filterPredicates));
+    } else {
+      predicates.add(criteriaBuilder.or(filterPredicates));
+    }
```
tlynema-bravolt, line 110: I know predicates is class-level.  Changing it on the fly like this makes this class hard to test.  Is there a better way (honestly, there may not be)?

src/main/java/com/bravo/user/dao/specification/UserSpecification.java
```diff
@@ -27,5 +30,13 @@ public void doFilter(
     applyStringFilter(root.get("firstName"), filter.getFirstNames());
     applyStringFilter(root.get("lastName"), filter.getLastNames());
     applyStringFilter(root.get("middleName"), filter.getMiddleNames());
+
+    if(StringUtils.hasText(filter.getName())){
```
tlynema-bravolt, line 34: This is a side-effect of using the same class to store the name.

src/main/java/com/bravo/user/validator/UserValidator.java
```diff
@@ -15,6 +15,16 @@ public void validateId(String id){
       throw new BadRequestException("'id' is required");
     }
   }
+  public void validateName(String name){
+    if(ValidatorUtil.isInvalid(name)){
+      throw new BadRequestException("'name' is required");
+    }
+    if(!name.matches("[A-Za-z]*")){
+      throw new BadRequestException("'name' cannot contain numbers, spaces, or special characters");
```
tlynema-bravolt, line 23: Like that you validated name separate from the other methods.  Should the greedy match be a ? here and not a * to avoid an empty string?

src/main/java/com/bravo/user/service/UserService.java
unamanic: This is an `info` in the existing methods.  Changing this to a debug should be done along with all of the others after a broader discussion on standards.

src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
unamanic: Every other method in the class adds predicates to this list which are then all anded together.  I'm not sure there is a much better way without refactoring the whole class.  In either event, tests could be accomplished by executing the methods and looking for the appropriate predicates in the list.

src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
```diff
@@ -79,28 +80,57 @@ protected void applyDateTimeFilter(
...
-      if(isLike){
-        predicate = criteriaBuilder.like(targetPath, targetValue.replace("*", "%"));
-      } else {
-        predicate = criteriaBuilder.equal(targetPath, targetValue);
-      }
-
-      if(isNot){
-        predicate = criteriaBuilder.not(predicate);
-      }
+      Predicate predicate = createPredicate(targetPath, targetValue, isLike, isNot);
```
wheelers-websites, line 91: You can move the result from `createPredicate` directly into the `predicates.add` method.

src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
```diff
@@ -79,28 +80,57 @@ protected void applyDateTimeFilter(
...
       predicates.add(predicate);
     }
     if(!inClause.isEmpty()){
       applyInFilter(targetPath, inClause);
     }
   }
+
+  protected void applyStringFilterToFields(Collection<Expression<String>> fields, String value) {
```
wheelers-websites, line 99: Is it possible to overload the `applyStringFilter` method instead of duplicating some of the code used in that method?

src/main/java/com/bravo/user/dao/specification/UserSpecification.java
wheelers-websites: I would've preferred to see this validation inside the `applyStringFilterToFields` method similar to how `applyStringFilter` is implemented so if the value provided is invalid the filter is not applied at all. I think that is cleaner than validation inside the Specification itself.

src/main/java/com/bravo/user/controller/UserController.java
unamanic: In practice I think Spring would reject the call if `name` were not passed since is a required parameter.  I'm not sure that we'd ever actually see an NPE here, but a null check would be a good idea.  I've added that along with moving the `replaceAll` method to validator utils.

src/main/java/com/bravo/user/model/filter/UserFilter.java
unamanic: I've created a `UserNameFuzzyFilter` and `UserNameFuzzySpecification` to prevent inadvertent use of name fuzzy matching when exact name matching is desired.

src/main/java/com/bravo/user/dao/specification/UserSpecification.java
unamanic: I've created a UserNameFuzzyFilter and UserNameFuzzySpecification to prevent inadvertent use of name fuzzy matching when exact name matching is desired.

src/main/java/com/bravo/user/validator/UserValidator.java
unamanic: In practice, an empty string would not make it here, since the previous line would throw a "Name is Required" error.  A `?` will match exactly one, I've switched this to a `+` and added beginning and end of line characters so it expects the name to contain 1 or more alpha characters.

src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
unamanic: Moved

src/main/java/com/bravo/user/dao/specification/AbstractSpecification.java
unamanic: It is possible, but I'm not sure it would be cleaner, the extraction of `createPredicate` eliminated most of the duplicated code.  It would be good to breifly pair with you to get a better understanding of what you are thinking.

wheelers-websites (approved):

PR #7, opened 2021-08-30
15-0015-backend-login-functionality-for-user

danny-huertas:
Updated User DAO/DTO to include new auth fields
Created New POST API to Validate User Password
Updated Data Seed to Update Mock Data
Added some jUnit to test the code added

Generally I would have had a conversation with you before implementing these changes to make sure we were on the same page, but I tried my best to interpret the story requirements for what was needed and went from there.

I've included an ERD below (I didn't fully implement what I'm displaying in this for roles cause the story required an enum for ADMIN), but the way I've broken out the code, it would be easy to enhance roles to give user one to many logic on roles stored in the db as displayed in this ERD.

auth-erd

Below are the screenshots to ensure the App Login API is Functioning Locally. I also added the postman collection as well as environment file to the resource folder so whomever picks this up can pull the collection and run it locally.

user-get-by-name
auth-validate-valid-case-insensitive
auth-validate-valid-admin
auth-validate-invalid-password
auth-validate-invalid-email


src/main/java/com/bravo/user/dao/model/Auth.java
```diff
@@ -0,0 +1,33 @@
...
+  private String email;
+
+  @Column(nullable = false)
+  private String password;
+
+  @Enumerated(EnumType.STRING)
+  @Column
+  private Role role;
+
+  @OneToOne(cascade = CascadeType.ALL)
```
tlynema-bravolt, line 30: Like that you added a CascadeType.  Is CascadeType.PERSIST necessary in this case?

src/main/java/com/bravo/user/dao/model/User.java
```diff
@@ -3,21 +3,24 @@
...
-import lombok.Data;
+import javax.persistence.*;
+
+import lombok.*;
 
 @Entity
-@Data
+@Getter
+@Setter
+@AllArgsConstructor
```
tlynema-bravolt, line 13: With lombok not generating toString, Equals and Hashcode;  what is the desired effect from this change?

src/main/java/com/bravo/user/dao/model/mapper/ResourceMapper.java
```diff
@@ -22,15 +23,21 @@ public ResourceMapper(MapperFacade mapperFacade) {
...
+      String name;
+      if (user.getMiddleName() != null && !user.getMiddleName().trim().isEmpty()) {
+          name = String.format("%s %s %s", user.getFirstName(), user.getMiddleName(), user.getLastName());
+      } else {
+          name = String.format("%s %s", user.getFirstName(), user.getLastName());
+      }
+      dto.setName(name);
+
+      //add auth email and role to user dto
+      if (user.getAuth() != null) {
```
tlynema-bravolt, line 37: I agree with this approach.  This enables a user to not have to have an entry in the authentication table.  Is there a way in code to enforce every user to also have a login?

src/main/java/com/bravo/user/dao/model/Auth.java
danny-huertas: When I added it with ALL, I was thinking their might be use cases for CRUD functionality driven from the auth level. For example an admin, or an import job used to create auth users with default passwords.  I also added the constraint at the db level as seen in the screenshot, but depending on the use cause the CascadeType could be changed.
table-constraint

src/main/java/com/bravo/user/dao/model/User.java
danny-huertas: Generally, I only use the Data annotation in model objects cause it can cause performance and memory issues as seen in the screenshot below.  It's not really noticeable when the entities don't have any join constrains, but as entity relationships are built out, this can cause issues.  

I could have left the Data tag and added the ToString.Exclude tag to the OneToOne relationship I added for user and auth, but I generally use Getter and Setter along with the toString, Equals and Hashcode as needed on entities (It's a preference, but I'm fine with either approach)
data-tag

src/main/java/com/bravo/user/dao/model/mapper/ResourceMapper.java
danny-huertas: Honestly, with the db constraint I added that ties user to auth, I don't think this null check is needed, but I added it cause I wasn't sure if a null pointer exception would be thrown from any of the tests that build out the user entities.  Now that I think about it, I don't think an exception would have been thrown and I also could have added the NotNull annotation to role and email to enforce them at the user bean level and optional = false to the auth mapping in the user entity.

danny-huertas:
@tlynema-bravolt Hi Thomas
Thanks for the detailed feedback on this PR.  I replied to all your comments, but didn't make any changes to the code I submitted cause I wasn't sure what the process is on this assessment (If this was for real, I would have at least added optional = false to the OneToOne mapping of Auth in User).

tlynema-bravolt:
> @tlynema-bravolt Hi Thomas
> Thanks for the detailed feedback on this PR. I replied to all your comments, but didn't make any changes to the code I submitted cause I wasn't sure what the process is on this assessment (If this was for real, I would have at least added optional = false to the OneToOne mapping of Auth in User).

@danny-huertas it's all good.  Thanks for taking the time to show us your skills and replying to the comments.  Keep in touch with Bill :+1:

PR #14, opened 2021-11-09
Initial review for Wright

bravocw:


src/main/java/com/bravo/user/dao/model/Profile.java
```diff
@@ -0,0 +1,41 @@
...
+  @Column(name = "image_ref", nullable = false)
+  private String imageRef;
+
+  @Column(name = "updated", nullable = false)
+  private LocalDateTime updated;
+
+  public Profile() {
+    super();
+    this.id = UUID.randomUUID().toString();
+    this.updated = LocalDateTime.now();
```
tlynema-bravolt, line 33: Having the update time be when the object is created may not match the actual update time.

bravocw: This is true - I'll admit I went with the design in the other VO due to time. I'd expect the update time to either be populated immediately before the create/update action or via a trigger in the db.

src/main/java/com/bravo/user/dao/model/User.java
```diff
@@ -33,6 +40,10 @@
   @Column(name = "updated", nullable = false)
   private LocalDateTime updated;
 
+  @OneToOne(cascade = CascadeType.REMOVE, fetch = FetchType.LAZY)
```
tlynema-bravolt, line 43: Bonus points for adding the CascadeType.  Would like to know why this is lazily fetched.

bravocw: Lazy fetch due to discussion in email - agreed that the API for loading users would not be loading the profile entity as part of it. If API would need to load it together in the future, the query would need to call out the desire to fetch that profile.

PR #18, opened 2021-11-14
0015: Add user authentication.

dboggs95:
As you go through my code, you'll notice a few things:
1. I did research on industry standards set by OWASP and NIST. I used the recommended utility for the salt and the recommended algorithm for hashing, plus the recommended configurations for both. For example, I limited the password to 64 characters to prevent pre-hashing, which I read has some issues that aren't worth allowing more characters.
2. I took care that everywhere I handled a password, salt, or hash, I kept them in an array of primitives and wiped said array with spaces when I no longer needed them. I understand this is a common practice in Java to prevent copies of the password floating over all over memory and finding their way into a heap dump, and this is actually required for FIPS compliance in some use cases (like Digital Signing).
3. You'll also see I left references to the relevant online documentation throughout the code.

If you run my solution, I did a couple things you should know about:
1. I fixed the h2 console configuration so the URL would work. You will still have to enable it in the yml file though.
2. I managed to reconstruct the seed data will all the same users, but they now have e-mails, salts, and hashes.
3. The password for every user is "Bravo.Firstname.Lastname". For example: Joyce Lucas Roberts has the password, "Bravo.Joyce.Roberts".

There were some ambiguities I'd like to bring up:
1. I was not given coding standards. I just let Eclipse Formatter do its thing, so my code doesn't look the same as the pre-existing code. I follow coding standards religiously though. If I am given a formatter and a written set of rules, this won't be an issue.
2. All I was told to do was return a 200 from the authentication API. I wasn't sure what to put in the response body, so I just added the user data and a placeholder for a session token.
3. You might consider some of the logic around the additions of the e-mail and user role to be a little weak. Again, there weren't too many specifics about these. 
    - For example, I wasn't told to update the create, retrieve, and update API's, but they need to be. If I attempt to create a user, I currently have no way of telling that it fails because the user already existed (that is, the e-mail is already in use). 
    - As for the user role. Since tables are forever, I put the data first, and I created a table defining the valid values of the user role with text names and integer ID's for foreign keys. But in the code, I use an enum (like the card said) rather than creating a JPA entity to represent that table, since it wasn't required for this specific use case.


dboggs95:
Test 1 - Create a new user:
image

Test 2 - Retrieve this user's record:
image

Test 3 - Authenticate this user:
image

dboggs95:
Test 1 - Retrieve the ADMIN:
image

Test 2 - Authenticate admin credentials:
image

dboggs95:
Test 1 - Retrieve an existing user record:
image

Test 2 - Authenticate that user:
image

dboggs95:
Test 1 - Attempt to authenticate with bad credentials:
image

Test 2 - Attempt to authenticate with bad e-mail:
image

Test 3 - Attempt to authenticate with valid e-mail, not in our records:
image

dboggs95:
Test 1 - Attempt to create a user with a bad e-mail:
image

Test 2 - Attempt to create a user with a bad password, too short:
image

Test 3 - Attempt to create a user with a bad password, too long:
image

PR #19, opened 2021-11-15
Added tests specified by ticket 0008

alecStewart1:


alecStewart1:
Been 2 weeks and there's no reveiw. Let me know if it should be opened up again.

PR #22, opened 2021-11-24
Update to candidate-java-sample from David Valen

dvalen14:
Here's my updates to the sample java file.   I have the sample passwords in a spreadsheet I can share via email or other method.  Email: david.valen@live.com

Thank you for the opportunity.  Hope to hear from you soon.


src/main/java/com/bravo/user/service/UserService.java
```diff
@@ -49,7 +58,22 @@ public UserReadDto retrieve(final String id){
     LOGGER.info("found user '{}'", id);
     return resourceMapper.convertUser(user);
   }
-
+  
+  public PasswordValidateDto validate(final PasswordValidateDto passwordValidateDto, final PageRequest pageRequest, final HttpServletResponse httpResponse){
+      final String encryptedPassword = encryptService.encrypt(passwordValidateDto.getPassword());
+      User user = userRepository.findByEmail(passwordValidateDto.getEmail());
+      
+      if(!passwordValidateDto.getEmail().matches("^([_a-zA-Z0-9-]+(\\.[_a-zA-Z0-9-]+)*@[a-zA-Z0-9-]+(\\.[a-zA-Z0-9-]+)*(\\.[a-zA-Z]{1,6}))?$")){
```
tlynema-bravolt, line 66: Could this use the isEmailValid below?

src/main/java/com/bravo/user/utility/ValidatorUtil.java
```diff
@@ -46,6 +46,14 @@ else if(value instanceof String){
     }
     return isValid;
   }
+  
+  public static boolean isEmailValid(final String string){
+    return string != null && string.matches("^([_a-zA-Z0-9-]+(\\.[_a-zA-Z0-9-]+)*@[a-zA-Z0-9-]+(\\.[a-zA-Z0-9-]+)*(\\.[a-zA-Z]{1,6}))?$");
```
tlynema-bravolt, line 51: string is a hard to follow name here.  The method says to pass in an email so it's followable, inside of the method, we can't tell that string is and email.

src/main/java/com/bravo/user/service/UserService.java
```diff
@@ -49,7 +58,22 @@ public UserReadDto retrieve(final String id){
...
+      	throw new UserNamePasswordException("Invalid Email Format");
+      }
+      
+      if((user == null) || (!encryptedPassword.equals(user.getPassword()))){
```
tlynema-bravolt, line 70: Can this be done in a way that avoids a possible null return?

src/main/java/com/bravo/user/exception/UserNamePasswordException.java
```diff
@@ -0,0 +1,12 @@
+package com.bravo.user.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(HttpStatus.UNAUTHORIZED)
+public class UserNamePasswordException extends RuntimeException {
+
+  public UserNamePasswordException(final String message){
+    super(String.format("UserName/Passworod Invalid | %s", message));
```
DaveAldon, line 10: Simple misspelling here 😄

DaveAldon:
Would like to mention, we're not expecting any further commits in response to our comments. But you're invited to elaborate/comment on any of our questions. Thank you for your PR!

src/main/java/com/bravo/user/service/UserService.java
dvalen14: Agree...this should have used the isEmailValid from ValidatorUtil

if(ValidatorUtil.isEmailInvalid(passwordValidateDto.getEmail())){...

src/main/java/com/bravo/user/utility/ValidatorUtil.java
dvalen14: Again...agree here.   The variable should be more descriptive probably using "email" for the string variable name.

src/main/java/com/bravo/user/exception/UserNamePasswordException.java
dvalen14: Agree...would update...

src/main/java/com/bravo/user/service/UserService.java
dvalen14: Yes...using the included Lombok library, I should have done something like this:
      if(!(isPasswordValid(user, encryptedPassword)){
      	throw new UserNamePasswordException("Invalid Email/Password combination");
      }  
 
 private boolean isPasswordValid(@NonNull User user, @NonNull String encryptedPassword){
    return encryptedPassword.equals(user.getPassword());
  }

dvalen14:
Thanks for the feedback.   I'm always learning.  At the very least, I'll use this in future projects.

PR #23, opened 2021-12-21
Kyle Krueger - Add Address Retrieve Logic

kkrueger-bravo:
Testing Evidence - API Call
image

Testing Evidence - Logs
image


src/test/java/com/bravo/user/service/AddressServiceTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+		this.dtoAddresses = ids.stream().map(id -> createAddressDto(Integer.toString(id)))
+				.collect(Collectors.toList());
+
+		when(resourceMapper.convertAddresses(daoAddresses)).thenReturn(dtoAddresses);
+	}
+
+	@Test
+	void retrieveByUserId() {
+		String userId = "123a-456b";
```
wheelers-websites, line 60: `userId` could be `final` here.

src/test/java/com/bravo/user/controller/AddressControllerTest.java
```diff
@@ -0,0 +1,85 @@
...
+		String userId = "123a-456b";
+
+		when(addressService.retrieveByUserId(anyString())).thenReturn(addresses);
+
+		final ResultActions result = this.mockMvc.perform(get("/address/retrieve/".concat(userId)))
+				.andExpect(status().isOk());
+
+		for (int i = 0; i < addresses.size(); i++) {
+			result.andExpect(
+					jsonPath(String.format("$[%d].id", i)).value(addresses.get(i).getId()));
```
wheelers-websites, line 63: Weird formatting here.

src/main/java/com/bravo/user/controller/AddressController.java
```diff
@@ -0,0 +1,40 @@
...
+	public AddressController(AddressService addressService, UserValidator userValidator) {
+		this.addressService = addressService;
+		this.userValidator = userValidator;
+	}
+
+	@GetMapping(value = "/retrieve/{userId}")
+	@ResponseBody
+	public List<AddressDto> retrieve(final @PathVariable String userId,
+			final @RequestParam(required = false) Integer page,
+			final @RequestParam(required = false) Integer size,
```
wheelers-websites, line 34: You're accepting the `page` and `size` parameters, but not using them.

src/main/java/com/bravo/user/service/AddressService.java
```diff
@@ -0,0 +1,33 @@
...
+	private final AddressRepository addressRepository;
+	private final ResourceMapper resourceMapper;
+
+	public AddressService(AddressRepository addressRepository, ResourceMapper resourceMapper) {
+		this.addressRepository = addressRepository;
+		this.resourceMapper = resourceMapper;
+	}
+
+	public List<AddressDto> retrieveByUserId(final String userId) {
+		List<Address> addressList = addressRepository.findByUserId(userId);
```
wheelers-websites, line 28: `addressList` can be `final` here.

src/test/java/com/bravo/user/controller/AddressControllerTest.java
```diff
@@ -0,0 +1,85 @@
...
+	public void beforeEach() {
+		final List<Integer> ids = IntStream.range(1, 10).boxed().collect(Collectors.toList());
+
+		this.addresses = ids.stream().map(id -> createAddressDto(Integer.toString(id)))
+				.collect(Collectors.toList());
+	}
+
+	@Test
+	void getRetrieveByUserId() throws Exception {
+		String userId = "123a-456b";
```
wheelers-websites, line 54: `userId` can be `final` here.

src/main/java/com/bravo/user/service/AddressService.java
tlynema-bravolt: Does final on a list make it immutable -- for Kyle?

wheelers-websites:

src/test/java/com/bravo/user/dao/model/mapper/ResourceMapperTest.java
```diff
@@ -1,30 +1,26 @@
...
+import com.bravo.user.dao.model.Address;
+import com.bravo.user.dao.model.User;
+import com.bravo.user.model.dto.AddressDto;
+import com.bravo.user.model.dto.UserReadDto;
 
 @ContextConfiguration(classes = {App.class})
 @ExtendWith(SpringExtension.class)
 @SpringBootTest
-public class ResourceMapperTest {
+class ResourceMapperTest {
```
tlynema-bravolt, line 23: Bonus points for this :)

src/test/java/com/bravo/user/service/AddressServiceTest.java
```diff
@@ -0,0 +1,79 @@
...
+	private ResourceMapper resourceMapper;
+
+	@MockBean
+	private AddressRepository addressRepository;
+
+	private List<AddressDto> dtoAddresses;
+
+	@BeforeEach
+	public void beforeEach() {
+		final List<Integer> ids = IntStream.range(1, 10).boxed().collect(Collectors.toList());
```
tlynema-bravolt, line 45: I like streams.  This falls as a non-judicious use IMHO https://dev.to/kylec32/effective-java-use-stream-judiciously-4178.

src/main/java/com/bravo/user/controller/AddressController.java
kkrueger-bravo: good catch, I initially was going to implement pagination until I re-read the requirements to simply return all addresses for the userId. I will clean them up, as well as the unused httpResponse parameter

src/main/java/com/bravo/user/service/AddressService.java
kkrueger-bravo: @wheelerswebservices - in my past experience, I haven't been so liberal with specifying variables final, but after some research it sounds like this is beneficial in a few ways. IDEs are even able to implement missing keywords automatically as a Save action- thanks for the advice

@tlynema-bravolt - no, making a List (or any object) final does not make it's contents immutable, it simply locks in the reference

src/test/java/com/bravo/user/controller/AddressControllerTest.java
kkrueger-bravo: thanks- will fix

src/test/java/com/bravo/user/controller/AddressControllerTest.java
kkrueger-bravo: this was initially auto-formatted via Eclipse, but I will manually adjust this to make it single-lined to improve readability

src/test/java/com/bravo/user/service/AddressServiceTest.java
kkrueger-bravo: My goal was to remain as similar to the development and test approach to the User components (which is using this already). In order to adjust, I'm curious what are your thoughts though on why this is non-judicious? Is it not explicit enough or should it just be handled in a more straightforward way?

Would you recommend I hardcode a list of ids to test with instead?

src/test/java/com/bravo/user/service/AddressServiceTest.java
kkrueger-bravo: thanks- will fix

src/test/java/com/bravo/user/service/AddressServiceTest.java
tlynema-bravolt: > Would you recommend I hardcode a list of ids to test with instead?

Was thinking about this too.  A simple for loop would work here with a list.add.  The end result is to get the same List of integers and this code does that before every test.  This could be run once and used for all tests that way. 

I googled it to learn this one myself :) https://stackoverflow.com/questions/10242380/how-can-i-generate-a-list-or-array-of-sequential-integers-in-java  has variations with performance implications.  In this case, I think both of the lines could be combined into IntStream.rangeClosed(1, 10).forEach(x -> logic)  where the logic in the next line would be  within the stream.  That has the benefit of the range not being stored in a collection which is a bit better on memory.  

The code in the codebase is sometimes not perfect :) .   One of the goals is to see things that you pick up that we miss.  You did this in the test code which was awesome!

TLDR instead of stream -> List -> stream  go with just stream for the logic in lines 46-49.

PR #24, opened 2022-02-21
Add address retrieve API

HenryDev:
Address Retrieve API should return all data for a given User:
Screen Shot 2022-02-20 at 9 36 40 PM

### DIFF
--- a/src/main/java/com/bravo/user/dao/repository/AddressRepository.java
+++ b/src/main/java/com/bravo/user/dao/repository/AddressRepository.java
@@ -1,0 +1,10 @@
+package com.bravo.user.dao.repository;
+
+import com.bravo.user.dao.model.Address;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface AddressRepository extends JpaRepository<Address, String>, JpaSpecificationExecutor<Address> {
+}
--- a/src/main/java/com/bravo/user/dao/specification/AddressSpecification.java
+++ b/src/main/java/com/bravo/user/dao/specification/AddressSpecification.java
@@ -1,0 +1,19 @@
+package com.bravo.user.dao.specification;
+
+import com.bravo.user.dao.model.Address;
+import com.bravo.user.model.filter.AddressFilter;
+import lombok.RequiredArgsConstructor;
+
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Root;
+
+@RequiredArgsConstructor
+public class AddressSpecification extends AbstractSpecification<Address> {
+    private final AddressFilter filter;
+
+    @Override
+    void doFilter(Root<Address> root, CriteriaQuery<?> criteriaQuery, CriteriaBuilder criteriaBuilder) {
+        applyStringFilter(root.get("userId"), filter.getUserId());
+    }
+}
--- a/src/main/java/com/bravo/user/service/AddressService.java
+++ b/src/main/java/com/bravo/user/service/AddressService.java
@@ -1,0 +1,29 @@
+package com.bravo.user.service;
+
+import com.bravo.user.dao.model.Address;
+import com.bravo.user.dao.model.mapper.ResourceMapper;
+import com.bravo.user.dao.repository.AddressRepository;
+import com.bravo.user.dao.specification.AddressSpecification;
+import com.bravo.user.model.dto.AddressDto;
+import com.bravo.user.model.filter.AddressFilter;
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+import java.util.Set;
+
+@RequiredArgsConstructor
+@Service
+public class AddressService {
+    private final AddressRepository addressRepository;
+    private final ResourceMapper resourceMapper;
+
+    public List<AddressDto> findAllByUserId(String userId) {
+        AddressFilter filter = AddressFilter.builder()
+                .userId(Set.of(userId))
+                .build();
+        AddressSpecification specification = new AddressSpecification(filter);
+        List<Address> addresses = addressRepository.findAll(specification);
+        return resourceMapper.convertAddresses(addresses);
+    }
+}
--- a/src/test/java/com/bravo/user/controller/AddressControllerTest.java
+++ b/src/test/java/com/bravo/user/controller/AddressControllerTest.java
@@ -1,0 +1,49 @@
+package com.bravo.user.controller;
+
+import com.bravo.user.model.dto.AddressDto;
+import com.bravo.user.service.AddressService;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.web.servlet.MockMvc;
+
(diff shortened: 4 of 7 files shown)


src/main/java/com/bravo/user/dao/repository/AddressRepository.java
```diff
@@ -0,0 +1,10 @@
+package com.bravo.user.dao.repository;
+
+import com.bravo.user.dao.model.Address;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface AddressRepository extends JpaRepository<Address, String>, JpaSpecificationExecutor<Address> {
```
wheelerswebservices, line 9: `JpaSpecificationExecutor` is not really required here since we're only querying on a single input.

     List<Address> findByUserId(final String userId);

https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.core-concepts

src/main/java/com/bravo/user/dao/specification/AddressSpecification.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.bravo.user.dao.model.Address;
+import com.bravo.user.model.filter.AddressFilter;
+import lombok.RequiredArgsConstructor;
+
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Root;
+
+@RequiredArgsConstructor
```
wheelerswebservices, line 11: Why are you using `@RequiredArgsConstructor` instead of `@AllArgsConstructor`?

src/test/java/com/bravo/user/controller/AddressControllerTest.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@WebMvcTest(controllers = AddressController.class)
+class AddressControllerTest {
+    @Autowired
+    private MockMvc mockMvc;
+    @MockBean
+    AddressService service;
```
wheelerswebservices, line 23: `service` should be `private`.

src/test/java/com/bravo/user/service/AddressServiceTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Test
+    void findAllShouldCallRepositoryMethodAndMatchZip() {
+        Address address = new Address();
+        String zip = "test zip";
+        address.setZip(zip);
+        when(repository.findAll(any(Specification.class))).thenReturn(List.of(address));
+        List<AddressDto> results = testObj.findAllByUserId("");
+        assertEquals(zip, results.get(0).getZip());
+        verify(repository).findAll(any(Specification.class));
```
wheelerswebservices, line 45: You should not `verify` with concepts like `any()` since the implementation can become flawed and your test won't catch that (because you're mocking the `findAll`. Instead you should only use `any()` in your `when` statements and use the actual objects in your `verify` statements. If you do this then you can be less specific in your `when` statements.

     final String userId = "";
     // mock data     

     when(repository.findAll(any()).thenReturn(List.of(address));
   
     final List<AddressDto> results = addressService.findAllByUserId(userId);
     // asserts 

     final AddressFilter filter = new AddressFilter(Set.of(userId));
     final AddressSpecification specification = new AddressSpecification(filter);
     verify(repository).findAll(specification);

src/main/java/com/bravo/user/dao/repository/AddressRepository.java
HenryDev: Oh yeah that's right. I'm rusty there it seems. Last time I had to add a method to a repository interface was in 2015, or 7 years ago

src/main/java/com/bravo/user/dao/specification/AddressSpecification.java
HenryDev: I just use it by default, not much of a difference here I don't think. In case some one comes along and forgets to make their fields `final` maybe this can sorta catch that.

src/main/java/com/bravo/user/service/AddressService.java
```diff
@@ -0,0 +1,29 @@
...
+public class AddressService {
+    private final AddressRepository addressRepository;
+    private final ResourceMapper resourceMapper;
+
+    public List<AddressDto> findAllByUserId(String userId) {
+        AddressFilter filter = AddressFilter.builder()
+                .userId(Set.of(userId))
+                .build();
+        AddressSpecification specification = new AddressSpecification(filter);
+        List<Address> addresses = addressRepository.findAll(specification);
```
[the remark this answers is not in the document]
HenryDev: Yeah I used to have that generator box checked too, but nowadays I think if someone have bugs from reassigning variables, then they probably have a bigger problem. Generally variables don't ever get reassigned anyway so there's no need to have `final` all over the place. I think just having them on property fields should be enough, maybe occasionally parameters if it's some logic heavy method.

src/test/java/com/bravo/user/controller/AddressControllerTest.java
HenryDev: Agree.

src/test/java/com/bravo/user/controller/AddressControllerTest.java
```diff
@@ -0,0 +1,49 @@
...
+                .andExpect(jsonPath("$[1].zip").value("test zip 1"));
+    }
+
+    @Test
+    void emptyIdShouldNotFindAnything() throws Exception {
+        mockMvc.perform(get("/user/address")).andExpect(status().isNotFound());
+    }
+
+    @Test
+    void noDataShouldReturnEmptyList() throws Exception {
```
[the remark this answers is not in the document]
HenryDev: I typically use subjectMethodNameShouldBehaveThisWay as a pattern, so in this case it's enough, and yes if more endpoints get in here then it'll have to be updated.

src/test/java/com/bravo/user/service/AddressServiceTest.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.mockito.Mockito.when;
+
+@SpringBootTest
+@ExtendWith(MockitoExtension.class)
+class AddressServiceTest {
+    private AddressService testObj;
+    @Mock
+    private AddressRepository repository;
+    @Autowired
+    private ResourceMapper mapper;
```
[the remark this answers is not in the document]
HenryDev: Sometimes a field gets mocked but don't get called in a `when()`, then intellij says that variable is unused, which is not true but it messes with you when you're reading it.

With the mapper I think it being a utility component it's fine to have additional coverage on it, also gets rid of the need to do another `when()`

I call all subjectUnderTest `testObj` so that way it's quick and easy to see what method is the test about. I usually middle click on it to bring the cursor to the type, or by looking at the test file name or the class name.

https://stackoverflow.com/questions/7321407/what-is-sut-and-where-did-it-come-from

Keeping the test object name consistent makes it easy to spot in any test classes. I have all my tests' SUT named testObj, and the reason for having all tests using the same test object name is because all the tests always have the same purpose of running the SUT method and check results. The business logic varies, but tests' purpose and flow stay the same.

src/test/java/com/bravo/user/service/AddressServiceTest.java
HenryDev: Makes sense. Sometimes I tack it on the back of assertions just to make sure the method gets called but I guess it's not very useful in this case.

PR #25, opened 2022-03-17
[bravo-0013] Implement new payment API to receive payments by user id

lystive:
Ticket [bravo-0013](https://trello.com/c/avi8pE7N/13-0013-backend-add-payment-method-retrieve-api)

What was changed:

1) Added new payments API to retrieve payments by user id
2) Added new mapstruct dependency, since the existing dependency is useful is some cases
3) Update pom.xml
4 Additional fixes
5) Cover changes by Junit and Integration tests

Why I decided to use userId as Request Parameter?

For me it looks weird we would have in path variable, since then endpoint should looks like /users/{userId}/payments.
As was mentioned in ticket, we need to use PaymentController, so the best practices would be using userId as request param.
Sometimes it depends on what we would have in the future.

Also there was a couple of things:
We could fetch payments by creating new repository interface, as I've done, but on the other hand the implementation could be:
1) Simply create new method in UserController with endpoint /users/{userId}/payments 
2) Call user service and fetch user by id
3) Get List of payments of user if one exists.

But as I said it depends on project needs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,6 +16,15 @@
 
   <properties>
     <java.version>11</java.version>
+    <org.mapstruct.version>1.4.2.Final</org.mapstruct.version>
+    <guava.version>30.1.1-jre</guava.version>
+    <junit.jupiter.version>5.7.2</junit.jupiter.version>
+    <lombok.version>1.18.20</lombok.version>
+    <mysql.socket.factory.version>1.3.1</mysql.socket.factory.version>
+    <mysql.connector.version>8.0.24</mysql.connector.version>
+    <orika.core.version>1.5.4</orika.core.version>
+    <springfox.swagger2.version>2.5.0</springfox.swagger2.version>
+    <springfox.swagger.ui.version>2.5.0</springfox.swagger.ui.version>
   </properties>
 
 
@@ -32,7 +41,7 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>30.1.1-jre</version>
+      <version>${guava.version}</version>
     </dependency>
 
     <!-- Jackson -->
@@ -53,7 +62,7 @@
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-engine</artifactId>
-      <version>5.7.2</version>
+      <version>${junit.jupiter.version}</version>
       <scope>test</scope>
     </dependency>
 
@@ -61,7 +70,7 @@
     <dependency>
       <groupId>org.projectlombok</groupId>
       <artifactId>lombok</artifactId>
-      <version>1.18.20</version>
+      <version>${lombok.version}</version>
       <scope>provided</scope>
     </dependency>
 
@@ -69,19 +78,19 @@
     <dependency>
       <groupId>com.google.cloud.sql</groupId>
       <artifactId>mysql-socket-factory-connector-j-8</artifactId>
-      <version>1.3.1</version>
+      <version>${mysql.socket.factory.version}</version>
     </dependency>
     <dependency>
       <groupId>mysql</groupId>
       <artifactId>mysql-connector-java</artifactId>
-      <version>8.0.24</version>
+      <version>${mysql.connector.version}</version>
     </dependency>
 
     <!-- orika -->
     <dependency>
       <groupId>ma.glasnost.orika</groupId>
       <artifactId>orika-core</artifactId>
-      <version>1.5.4</version>
+      <version>${orika.core.version}</version>
     </dependency>
 
     <!-- Spring Boot -->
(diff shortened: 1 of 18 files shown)


wheelerswebservices:

pom.xml
```diff
@@ -16,6 +16,15 @@
...
     <java.version>11</java.version>
+    <org.mapstruct.version>1.4.2.Final</org.mapstruct.version>
+    <guava.version>30.1.1-jre</guava.version>
+    <junit.jupiter.version>5.7.2</junit.jupiter.version>
+    <lombok.version>1.18.20</lombok.version>
+    <mysql.socket.factory.version>1.3.1</mysql.socket.factory.version>
+    <mysql.connector.version>8.0.24</mysql.connector.version>
+    <orika.core.version>1.5.4</orika.core.version>
+    <springfox.swagger2.version>2.5.0</springfox.swagger2.version>
+    <springfox.swagger.ui.version>2.5.0</springfox.swagger.ui.version>
```
wheelerswebservices, line 27: Nice clean up here.

src/main/java/com/bravo/user/controller/advice/ExceptionHandlerAdvice.java
```diff
@@ -31,4 +34,17 @@ public ErrorDto handleBindException(final BindException exception){
     response.setStatusCode(400);
     return response;
   }
+
+  @ExceptionHandler(value = PaymentNotFoundException.class)
+  @ResponseBody
+  @ResponseStatus(value = HttpStatus.NOT_FOUND)
+  public ErrorDto handlePaymentNotFoundException(final PaymentNotFoundException exception){
+    log.info("Payments not found, details: {}", exception.getMessage());
+    var errorDto = new ErrorDto();
```
wheelerswebservices, line 43: `errorDto` could be `final`

src/main/java/com/bravo/user/dao/model/mapper/PaymentMapper.java
```diff
@@ -0,0 +1,23 @@
+package com.bravo.user.dao.model.mapper;
+
+import com.bravo.user.dao.model.Payment;
+import com.bravo.user.model.dto.PaymentDto;
+import java.util.List;
+import org.mapstruct.Mapper;
+import org.mapstruct.Mapping;
+import org.mapstruct.Named;
+
+@Mapper(componentModel = "spring")
```
wheelerswebservices, line 10: Can you explain what `componentModel = "spring"` means?

src/main/java/com/bravo/user/dao/model/mapper/ResourceMapper.java
```diff
@@ -41,17 +39,6 @@ public AddressDto convertAddress(final Address address){
     return dto;
   }
 
-  public <T extends Collection<Payment>> List<PaymentDto> convertPayments(final T payments){
-    return payments.stream().map(this::convertPayment).collect(Collectors.toList());
-  }
-
-  public PaymentDto convertPayment(final Payment payment){
```
wheelerswebservices, line 48: Can you elaborate on why the new `PaymentMapper` class is superior to these methods within the existing `ResourceMapper` class?

src/main/java/com/bravo/user/exception/PaymentNotFoundException.java
```diff
@@ -0,0 +1,12 @@
+package com.bravo.user.exception;
+
+import java.text.MessageFormat;
+
+public class PaymentNotFoundException extends RuntimeException {
+
+    private static final String MESSAGE = "Payment not found for user with id {0}";
+
+    public PaymentNotFoundException(String userId) {
+        super(MessageFormat.format(MESSAGE, userId));
```
wheelerswebservices, line 10: 1. Why did you use `MessageFormat.format` instead of `String.format`? 
2. Do you think this `MESSAGE` variable is really necessary as it's only used in 1 place?

src/main/java/com/bravo/user/dao/model/mapper/PaymentMapper.java
lystive: @wheelerswebservices it means construct mapper as spring bean. Because mapper has different construction types, using factory, spring beans etc.

src/main/java/com/bravo/user/dao/model/mapper/ResourceMapper.java
lystive: @wheelerswebservices because mapperFacade, which is used here is not the flexible one from SOLID principles perpective.
First of all, if you would have 10 dto's which you need to map from the entity its not a good idea to create 20 methods inside ResourseMapper class.
Second, is that you need to implement all of the extra mapping, or when field names are different. In the Mapstruct its simple, you only need an interface, and specify everything only using annotations. Its very useful and simple to understand. Its flexible when you need some extra mappings, for example as was for card mask. So everything for this mapping is located only in 1 Java interface.
Thats why mapstruct is so popular today.
You could read all the details below: https://www.baeldung.com/java-performance-mapping-frameworks

src/main/java/com/bravo/user/exception/PaymentNotFoundException.java
lystive: @wheelerswebservices regarding first: I don't think this is really matters
second: this exception is only related to Payment not found flow, so there is not reason to extract this message somewhere else.

src/main/java/com/bravo/user/model/dto/PaymentDto.java
```diff
@@ -1,13 +1,18 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
 import lombok.Data;
+import lombok.NoArgsConstructor;
 
 @Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
 public class PaymentDto {
```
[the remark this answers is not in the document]
lystive: Builder used to simplify construction for testing. Since we use Builder, NoArgs and AllArgs constrcustors is required for MapStruct.

src/main/java/com/bravo/user/controller/advice/ExceptionHandlerAdvice.java
lystive: Agree, thanks

PR #28, opened 2022-05-18
jaxtell - added get payment by user id functionality

jaxtell:


src/main/java/com/bravo/user/controller/PaymentController.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.web.bind.annotation.*;
+
+import com.bravo.user.annotation.SwaggerController;
+import com.bravo.user.model.dto.PaymentDto;
+import com.bravo.user.service.PaymentService;
+import com.bravo.user.validator.UserValidator;
+
+import javax.servlet.http.HttpServletResponse;
+
+@RequestMapping(value = "/payment")
```
vandent6, line 16: Any comment on how the endpoint should look if you were to design it without pattern following?

jaxtell: It depends on how it would be used.  If user id is always going to be present it might make sense to be /user/{userId}/payment.  If there are future requests that allow searching across users then I'd leave it top level.  Instead of using words in the URL I'd favor http verbs to indicate action, e.g. GET on /payment/user/{userId} to get all payments for a user and/or GET on /payment with parameters to get payment(s) based on query parameters, PUT on /payment to add a payment, PATCH on /payment to update a payment, POSTs on /payment/[ACTION] for more complex operations.  I'd want to balance keeping the URL short while still being descriptive of what it does.

PR #29, opened 2022-07-28
Implementation of Payment Retrieve API

EricRybarczyk:
This work covers Trello ticket 0013 - [backend] add payment method retrieve api.

It adds Spring components of PaymentController, PaymentService, and PaymentRepository to implement the Payment Retrieve GET API.  Unit tests are included for the controller and service implementations.

I also resolved what I believe was a bug in the ResourceMapper. Previously, the mapper was setting five characters into the PaymentDto.CardNumberLast4 field. I modified it to set four characters, as implied by the field name. A parameterized unit test was added to ResourceMapperTest to verify this change.

Two screenshots from Postman are attached, showing example results from this API call running in a local environment. This uses sample data I added to the existing SQL script.

Please let me know if you have any feedback. Thanks!

Bravo-GetPaymentsApi-example1
Bravo-GetPaymentsApi-example2


src/test/java/com/bravo/user/service/PaymentServiceTest.java
```diff
@@ -0,0 +1,84 @@
...
+    private PaymentRepository paymentRepository;
+
+    private List<PaymentDto> dtoPayments;
+
+    @BeforeEach
+    void setUp() {
+        final List<Integer> ids = IntStream
+                .range(1, 10)
+                .boxed()
+                .collect(Collectors.toList());
```
tlynema-bravolt, line 46: Nitpicking here. Is this better than a for loop?

EricRybarczyk: I don't think you're nitpicking, and I'm glad you made me look more closely at this part of the code. My short answer is no, it's probably not better than a for loop - depending on what we actually do in the for loop I suppose.

Had I used a for loop I doubt I would have ended up with as much object creation as is happening in this code. The boxing of int primitives and even having the List<Integer> itself probably isn't the best for memory usage and execution time. On top of that, in the second stream operation creating the `paymentDtoList`, the call to `Integer.toString()` is passed the Integer object, so it gets unboxed again in that call. This could compound as the code base grows and more tests are added, since this is in the `@BeforeEach` setup.

So... I'd either go with a simple re-initialization and for loop like this:

```
paymentDtoList = new ArrayList<>();
for (int i = 0; i < 10; i++) {
    paymentDtoList.add(createPaymentDto(Integer.toString(i)));
}
```
        
Or at least a trimmed down stream operation like this:

```
paymentDtoList = IntStream
        .range(1, 10)
        .mapToObj(i -> createPaymentDto(Integer.toString(i)))
        .collect(Collectors.toList());
```
        
Being really honest, I saw that code in another test class in the project and I think my geek brain said "Oooh, I like streams"  and I used it without thinking critically enough about it.  This is one of the reasons that code reviews are so helpful!

Also, I'll nitpick on myself here... thinking through this also reminded me that the end parameter in `range(1, 10)` is exclusive, so you get a list of nine items, whereas my simple loop provides 10 items. It doesn't matter to these test results, but reading "1, 10" it is easy to mentally expect 10 items. We could use `rangeClosed(1, 10)` to get inclusive of the end param.

Thanks for the feedback and for questioning that code. Let me know if you'd like an update to the pull request to address the main concern with those stream operations.

PR #31, opened 2022-08-06
0013 - [backend] add payment method retrieve api

notoneword:
image
image

Working with and without paging parameters (see attached)

image
100% class test coverage

- Implemented solution
- fixed bugs with credit card number (showing last 5 digits, unique constraint)

### DIFF
--- a/src/main/java/com/bravo/user/model/dto/ErrorDto.java
+++ b/src/main/java/com/bravo/user/model/dto/ErrorDto.java
@@ -1,11 +1,12 @@
 package com.bravo.user.model.dto;
 
-import java.util.Set;
 import lombok.Data;
-import lombok.NoArgsConstructor;
+import lombok.RequiredArgsConstructor;
+
+import java.util.Set;
 
 @Data
-@NoArgsConstructor
+@RequiredArgsConstructor
 public class ErrorDto {
 
   private Set<String> errors;
--- a/src/main/java/com/bravo/user/service/PaymentService.java
+++ b/src/main/java/com/bravo/user/service/PaymentService.java
@@ -4,10 +4,14 @@ import com.bravo.user.dao.model.Payment;
 import com.bravo.user.dao.model.mapper.ResourceMapper;
 import com.bravo.user.dao.repository.PaymentRepository;
 import com.bravo.user.model.dto.PaymentDto;
+import com.bravo.user.utility.PageUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.PageRequest;
 import org.springframework.stereotype.Service;
 
+import javax.servlet.http.HttpServletResponse;
 import java.util.List;
 
 @Service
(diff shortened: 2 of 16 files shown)


src/main/java/com/bravo/user/service/PaymentService.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public List <PaymentDto> retrieveByUserId(
+          final String userId,
+          final PageRequest pageRequest,
+          final HttpServletResponse httpResponse
+  ){
+    final Page<Payment> paymentPage = paymentRepository.findPaymentsByUserId(userId, pageRequest);
+    final List<PaymentDto> payments = resourceMapper.convertPayments(paymentPage.getContent());
+    LOGGER.info("Returning {} payment(s) for user '{}'", payments.size(), userId);
+    LOGGER.trace("#### {} -- {}", pageRequest, payments);
```
tlynema-bravolt, line 38: Good use of string formatting here :+1:

src/main/resources/data.sql
```diff
@@ -18,7 +18,7 @@ create table address (
 create table payment (
     id varchar(60) primary key,
     user_id varchar(60) not null,
-    card_number varchar(16) not null unique,
+    card_number varchar(16) not null, -- No need to be unique (or only support payments by russian hackers using different card every time?)
```
tlynema-bravolt, line 21: varchar allows for characters.  Is that the right choice here?

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
```diff
@@ -0,0 +1,106 @@
...
+  private PaymentService paymentService;
+
+  private List<PaymentDto> payments;
+
+  @BeforeEach
+  public void beforeEach(){
+    final List<Integer> ids = IntStream
+        .range(1, 10)
+        .boxed()
+        .collect(Collectors.toList());
```
tlynema-bravolt, line 56: Nitpick here.  Why is this better than a simple for loop or a List.of?

src/main/java/com/bravo/user/model/dto/ErrorDto.java
```diff
@@ -1,11 +1,12 @@
...
-import java.util.Set;
 import lombok.Data;
-import lombok.NoArgsConstructor;
+import lombok.RequiredArgsConstructor;
+
+import java.util.Set;
 
 @Data
-@NoArgsConstructor
+@RequiredArgsConstructor
```
tlynema-bravolt, line 9: This change makes sense.  If making this change, would adding the final or @NonNull decorations also be needed?

src/main/java/com/bravo/user/validator/PaymentValidator.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.validation.Errors;
+
+@Component
+public class PaymentValidator extends CrudValidator {
+  public static final String USER_ID_MISSING_ERROR = "'User Id' in UUID format is required";
+
+  public void validateId(String id){
+    if(!ValidatorUtil.isUUIDValid(id)){
+      throw new BadRequestException(USER_ID_MISSING_ERROR);
+    }
```
tlynema-bravolt, line 15: The UUID object itself is immutable.  String is as well.  On the fence with this should return the UUID so it's potentially easier to code.  e.g. passing UUID, String into as method arguments vs String, String.  String, String may accidentally be flipped around.  Would like to know your thoughts on this one.

src/main/java/com/bravo/user/model/dto/ErrorDto.java
notoneword: good point, final modifier or @nonnull annotation is needed for required args.

src/main/resources/data.sql
notoneword: I think so.  The credit card 'number' is really a string consisting of numbers, you wouldn't want it to be treated like an actual number at any point, and you may do things to obscure part of the string, etc.

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
notoneword: Clean functional code, and was following conventions in other tests.

Even though a loop may be more performant at small numbers, I don't enough of a reason in a test class to make that change.

src/main/java/com/bravo/user/validator/PaymentValidator.java
notoneword: Lost the context a little with this question (not sure what method has two strings in signature currently?)  - but yeah, it could be cleaner to pass in UUID to validate method than a string.

notoneword:
thanks for the feedback & questions!

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
tlynema-bravolt: The other tests are wrong :) .  Bike shedding is a real thing so I very much see that viewpoint.  In this case, I do prefer List.of if only because the values are explicit, but there are usually more important things to do.

src/main/resources/data.sql
tlynema-bravolt: Thinking of it as a string of numbers is somewhere my mind didn't go initially.  I like it :) .  There are some numerical checks on credit cards like something that starts with a 4 is Visa and checks where the values can't be transposed.  That seems possible to determine if it is a char or a number so not an issue unless we are squeezing performance.  A number would likely save some DB space and a bit of memory as it would be able to be an unsigned number.

tlynema-bravolt:
Thanks for the comments as well :) .  It's always fun to see how people code and chat about solutions.

PR #36, opened 2022-11-21
0013 - [backend] add payment method retrieve api

parapetual:
- Add Payment Retrieve API (GET for a specific userId)
- Fix bug in ResourceMapper.convertPayment() extracting last 5 digits of card instead of 4
- Created com.bravo.payment package and moved Payment-specific classes there
- Renamed PaymentDto to PaymentReadDto - service will need PaymentSaveDto soon to actually store encrypted CC number
- Refactor: move classes shared between domains (user/payment) to new "common" package
- Left 2 TODOs that need attention (discuss with teammates, architects, etc.)


src/main/java/com/bravo/payment/controller/PaymentController.java
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Returns all payments made by a given user by their userId
+   */
+  @ApiOperation("Retrieve payments made by a user")
+  @GetMapping(value = "/retrieve/{userId}")
+  @ResponseBody
+  public List<PaymentReadDto> retrieve(
+      final @PathVariable String userId,
+      final @RequestParam(required = false) Integer page,
+      final @RequestParam(required = false) Integer size,
```
tlynema-bravolt, line 42: Is it ok to pass in nulls here?

src/main/java/com/bravo/payment/controller/PaymentController.java
```diff
@@ -0,0 +1,53 @@
...
+      final HttpServletResponse httpResponse
+  ) {
+    if (userId != null) {
```
tlynema-bravolt, line 45: userId should never be null,  I think :) Guessing though.

src/main/java/com/bravo/payment/model/dto/PaymentReadDto.java
```diff
@@ -1,10 +1,10 @@
-package com.bravo.user.model.dto;
+package com.bravo.payment.model.dto;
 
 import java.time.LocalDateTime;
 import lombok.Data;
 
 @Data
-public class PaymentDto {
+public class PaymentReadDto {
```
tlynema-bravolt, line 7: What makes a Dto a ReadDto?  This naming pattern is new one to me.

src/main/java/com/bravo/App.java
```diff
@@ -1,4 +1,4 @@
-package com.bravo.user;
+package com.bravo;
```
tlynema-bravolt, line 1: Like the rename here. :+1:

src/test/java/com/bravo/payment/service/PaymentServiceTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+  final String userId = randomUUID().toString();
+  private List<PaymentReadDto> dtoPayments;
+
+  @BeforeEach
+  public void beforeEach() {
+    final List<Integer> ids = IntStream
+        .range(1, 10)
+        .boxed()
+        .collect(Collectors.toList());
```
tlynema-bravolt, line 55: Is this final list still mutable?

src/main/resources/data.sql
```diff
@@ -648,3 +649,17 @@ insert into address (id, user_id, line1, line2, city, state, zip) values
 ('42f33d30-f3f8-4743-a94e-4db11fdb747d', '008a4215-0b1d-445e-b655-a964039cbb5a', '412 Maple St', null, 'Dowagiac', 'Michigan', '49047'),
 ('579872ec-46f8-46b5-b809-d0724d965f0e', '00963d9b-f884-485e-9455-fcf30c6ac379', '237 Mountain Ter', 'Apt 10', 'Odenville', 'Alabama', '35120'),
 ('95a983d0-ba0e-4f30-afb6-667d4724b253', '00963d9b-f884-485e-9455-fcf30c6ac379', '107 Annettes Ct', null, 'Aydlett', 'North Carolina', '27916');
+
+insert into payment (id, user_id, card_number, expiry_month, expiry_year) values
+-- Elian Gibson
+('877ca9e9-d28b-4889-a086-7167785395fe', 'fd6d21f6-f1c2-473d-8ed7-f3f9c7550cc9', '1111222233334444', 1, 2025),
+('6d8bb16f-28b1-469b-8013-2f5a01e4b8da', 'fd6d21f6-f1c2-473d-8ed7-f3f9c7550cc9', '1111222233334445', 2, 2025),
+('65826d22-44e9-459c-80da-103a1f576500', 'fd6d21f6-f1c2-473d-8ed7-f3f9c7550cc9', '1111222233334446', 3, 2025),
```
tlynema-bravolt, line 657: Nice sample data here.  It should hit some potential edge cases.

src/main/java/com/bravo/payment/controller/PaymentController.java
parapetual: Yep, PageUtil very much supports that case

src/main/java/com/bravo/payment/controller/PaymentController.java
parapetual: Good point, impossible scenario (and validator handles null checks too)

src/main/java/com/bravo/payment/model/dto/PaymentReadDto.java
parapetual: The linked story/requriements (0014) for this story seemed a bit inconsistent in its ask, but the intention very much seemed to be to store the encrypted full CC number (SaveDto), but read back only the last 4 (ReadDto).  This merits discussion with POs.

src/test/java/com/bravo/payment/service/PaymentServiceTest.java
parapetual: Yep. This was more habit, but also avoids the more likely case of reassigning the list by mistake.

src/main/java/com/bravo/payment/service/PaymentService.java
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public List<PaymentReadDto> retrieveByUserId(
+      final String userId,
+      final PageRequest pageRequest,
+      final HttpServletResponse httpResponse
+  ) {
+    final Page<Payment> paymentPage = paymentRepository.findByUserId(userId, pageRequest);
+    final List<PaymentReadDto> payments = resourceMapper.convertPayments(paymentPage.getContent());
+    LOGGER.info("found {} payment(s)", payments.size());
```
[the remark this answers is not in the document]
parapetual: The first thing that comes to mind is a debug-level log instead, but it should involve a discussion of TPS rates and such. Also, I would have a different concern if **either** this endpoint gets hit a lot **or** logging is _that_ expensive.
It is by no stretch noisy logging either (again, depends on use though).

PR #39, opened 2022-12-09
Robert test branch

rdsmart97:
#### Revision 2:
Refactoring PaymentController and PaymentControllerTest code to revmove redundant variables.

### Why Are We Making This Change?
- We want to be able to retrieve payment information for a user.  It's possible for a user to have multiple payment methods so all payment methods should be returned for the specified user.  

### What Is Changing In This Review?
- Added new controller for Payment APIs
- Added new service for Payment APIs
- Added API to retrieve payment by user Id
- Added test payment data to data.sql for local testing
- Changed substring length for ResourceMapper convert payment method because the old value was resulting in a string of 5 characters for "setCardNumberLast4".

##  How Was This Tested?
- Manually tested using swagger console: http://localhost:9999/swagger-ui.html
- Added new unit tests and ran them to make sure they all pass

### DIFF
--- a/src/main/java/com/bravo/user/controller/PaymentController.java
+++ b/src/main/java/com/bravo/user/controller/PaymentController.java
@@ -4,7 +4,6 @@ import com.bravo.user.annotation.SwaggerController;
 import com.bravo.user.model.dto.PaymentDto;
 import com.bravo.user.service.PaymentService;
 import com.bravo.user.validator.UserValidator;
-import lombok.NonNull;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -26,7 +25,7 @@ public class PaymentController {
 
 	@GetMapping(value = "/retrieve/{userId}")
 	@ResponseBody
-	public List<PaymentDto> retrieve(@NonNull final @PathVariable String userId) {
+	public List<PaymentDto> retrieve(final @PathVariable String userId) {
 		userValidator.validateId(userId);
 		return paymentService.retrieveByUserId(userId);
 	}
--- a/src/test/java/com/bravo/user/controller/PaymentControllerTest.java
+++ b/src/test/java/com/bravo/user/controller/PaymentControllerTest.java
@@ -42,9 +42,8 @@ class PaymentControllerTest {
 
 	@BeforeEach
 	public void beforeEach() {
-		final List<Integer> ids = IntStream.range(1, 10).boxed().collect(Collectors.toList());
-
-		this.payments = ids.stream().map(id -> createPaymentDto(Integer.toString(id)))
+		this.payments = IntStream.range(1, 10).boxed()
+				.map(id -> createPaymentDto(Integer.toString(id)))
 				.collect(Collectors.toList());
 	}
 


src/main/java/com/bravo/user/controller/PaymentController.java
```diff
@@ -0,0 +1,33 @@
...
+	private final UserValidator userValidator;
+
+	public PaymentController(PaymentService paymentService, UserValidator userValidator) {
+		this.paymentService = paymentService;
+		this.userValidator = userValidator;
+	}
+
+	@GetMapping(value = "/retrieve/{userId}")
+	@ResponseBody
+	public List<PaymentDto> retrieve(@NonNull final @PathVariable String userId) {
```
wheelerswebservices, line 29: What benefit does `@NonNull` provide here?

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
```diff
@@ -0,0 +1,82 @@
...
+	private PaymentService paymentService;
+
+	private List<PaymentDto> payments;
+
+	@BeforeEach
+	public void beforeEach() {
+		final List<Integer> ids = IntStream.range(1, 10).boxed().collect(Collectors.toList());
+
+		this.payments = ids.stream().map(id -> createPaymentDto(Integer.toString(id)))
+				.collect(Collectors.toList());
```
wheelerswebservices, line 48: `ids` variable is unnecessary since it's only used once. You can combine these statements together.

```
this.payments = IntStream.range(1, 10).boxed()
    .map(id -> createPaymentDto(Integer.toString(id)))
    .collect(Collectors.toList())
```

src/main/java/com/bravo/user/dao/model/mapper/ResourceMapper.java
```diff
@@ -48,7 +48,7 @@ public <T extends Collection<Payment>> List<PaymentDto> convertPayments(final T
   public PaymentDto convertPayment(final Payment payment){
     final String cardNumber = payment.getCardNumber();
     final PaymentDto dto = mapperFacade.map(payment, PaymentDto.class);
-    dto.setCardNumberLast4(cardNumber.substring(cardNumber.length() - 5));
+    dto.setCardNumberLast4(cardNumber.substring(cardNumber.length() - 4));
```
wheelerswebservices, line 51: Good job catching this bug! 👍🏻

src/main/java/com/bravo/user/controller/PaymentController.java
rdsmart97: I added the @NonNull annotation here to make sure that a NullPointerException would be thrown if this method is ever called without userId since userId is required for this method.

rdsmart97: I just tested this some more by making curl calls to the API on my terminal.  It seems like even if I don't provide a userId the API just treats this as a userId with the value of EMPTYSTRING so I think you are right and this check is not needed.  I think the only way userId will be null is if someone specifically passes in an encoded version of null and when I tested that, it was causing the API to throw a BadRequest exception so I think this is already handled.

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
rdsmart97: Good call out.  I have refactored the code and fixed this to removed the redundant variable.

PR #45, opened 2023-04-06
Added Payment Resource following User Resource conventions

dschroeter26:
-Added Payment Resource following User Resource conventions

Screenshot 2023-04-06 at 11 12 19 AM


src/test/java/com/bravo/user/controller/PaymentControllerTest.java
```diff
@@ -0,0 +1,58 @@
...
+        PaymentDto mockPaymentDtoOne = Mockito.mock(PaymentDto.class);
+        PaymentDto mockPaymentDtoTwo = Mockito.mock(PaymentDto.class);
+        List<PaymentDto> mockPaymentDtos = Arrays.asList(mockPaymentDtoOne, mockPaymentDtoTwo);
+        when(paymentService.retrieveByUserId(anyString(), isNull(), isA(HttpServletResponse.class)))
+                .thenReturn(mockPaymentDtos);
+
+        String userId = "008a4215-0b1d-445e-b655-a964039cbb5a";
+        final ResultActions result = this.mockMvc
+                .perform(get("/payment/retrieve/" + userId))
+                .andExpect(status().isOk());
```
tlynema-bravolt, line 53: This tests the happy path.  Are there other responses to check?

.gitignore
```diff
@@ -14,3 +14,5 @@ bin/
 
 # Log Files
 *.log
+
+.metadata/
```
wheelerswebservices, line 18: What IDE are you using to need to exclude `.metadata/`?

src/main/java/com/bravo/user/controller/PaymentController.java
```diff
@@ -0,0 +1,36 @@
...
+    @GetMapping(value = "/retrieve/{userId}")
+    @ResponseBody
+    public List<PaymentDto> retrieve(
+            final @PathVariable String userId,
+            final @RequestParam(required = false) Integer page,
+            final @RequestParam(required = false) Integer size,
+            final HttpServletResponse httpResponse
+    ) {
+            final PageRequest pageRequest = PageUtil.createPageRequest(page, size);
+            return paymentService.retrieveByUserId(userId, pageRequest, httpResponse);
```
wheelerswebservices, line 32: Is it safe to use the `userId` as is? 😨

src/main/java/com/bravo/user/dao/repository/PaymentRepository.java
```diff
@@ -0,0 +1,13 @@
...
+import com.bravo.user.dao.model.Payment;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
+
+import java.util.List;
+
+public interface PaymentRepository extends JpaRepository<Payment, String>, JpaSpecificationExecutor<Payment> {
+    Page<Payment> findAllByUserId(String userId, Pageable pageable);
```
wheelerswebservices, line 12: How is this different than `findByUserId`?  Is `JpaSpecificationExecutor` actually required here?

src/main/java/com/bravo/user/service/PaymentService.java
```diff
@@ -0,0 +1,42 @@
...
+        this.paymentRepository = paymentRepository;
+        this.resourceMapper = resourceMapper;
+    }
+
+    public List<PaymentDto> retrieveByUserId(
+            final String userId,
+            final PageRequest pageRequest,
+            final HttpServletResponse httpResponse
+    ){
+        Page<Payment> paymentPage = paymentRepository.findAllByUserId(userId, pageRequest);
```
wheelerswebservices, line 35: - `paymentPage` could be `final`. 
- Variable names can be a bit more generic if you consider the context of the service they are used within. e.g. `paymentPage` can become `page`, `response`, etc.

src/main/java/com/bravo/user/service/PaymentService.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.data.domain.PageRequest;
+import org.springframework.stereotype.Service;
+
+import javax.servlet.http.HttpServletResponse;
+import java.util.List;
+
+@Service
+public class PaymentService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(UserService.class);
```
wheelerswebservices, line 20: This `Logger` is bound to the wrong class.

src/main/resources/data.sql
```diff
@@ -648,3 +648,20 @@ insert into address (id, user_id, line1, line2, city, state, zip) values
...
+('0005', '008a4215-0b1d-445e-b655-a964039cbb5a', '6011123456789123', 5, 2025, NOW()),
+('0006', '008a4215-0b1d-445e-b655-a964039cbb5a', '5252803995123456', 6, 2026, NOW()),
+('0007', '008a4215-0b1d-445e-b655-a964039cbb5a', '6011012345678912', 7, 2027, NOW()),
+('0009', '008a4215-0b1d-445e-b655-a964039cbb5a', '3584587763245678', 9, 2024, NOW()),
+('0010', '008a4215-0b1d-445e-b655-a964039cbb5a', '6011000990139424', 10, 2025, NOW()),
+('0011', '008a4215-0b1d-445e-b655-a964039cbb5a', '4532016747591287', 11, 2026, NOW()),
+('0012', '008a4215-0b1d-445e-b655-a964039cbb5a', '6011514433546207', 12, 2027, NOW()),
+('0013', '008a4215-0b1d-445e-b655-a964039cbb5a', '6011313207071335', 1, 2023, NOW()),
+('0014', '008a4215-0b1d-445e-b655-a964039cbb5a', '4024007162397651', 2, 2024, NOW()),
+('0017', '008a4215-0b1d-445e-b655-a964039cbb5a', '6011514433546211', 6, 2027, NOW());
```
wheelerswebservices, line 666: - Any reason why you choose to use this `id` pattern? It seems problematic when you try to add the 10,000 record. 
- You should have added some different `user_id` values as well. Seems like this 1 user has way too many credit cards. 😏 
- `updated` has a `default` rule when excluded so you can achieve the same result by dropping the `updated` column altogether. note: `updated timestamp not null default current_timestamp()`.

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
```diff
@@ -0,0 +1,58 @@
...
+    private MockMvc mockMvc;
+
+    @MockBean
+    private PaymentService paymentService;
+
+    @Test
+    public void retrievePaymentWithUserIdAndNoPagination() throws Exception {
+        PaymentDto mockPaymentDtoOne = Mockito.mock(PaymentDto.class);
+        PaymentDto mockPaymentDtoTwo = Mockito.mock(PaymentDto.class);
+        List<PaymentDto> mockPaymentDtos = Arrays.asList(mockPaymentDtoOne, mockPaymentDtoTwo);
```
wheelerswebservices, line 46: - I don't think it's wise to `mock` the `PaymentDto` here. You should return real instances with mock data and use something like `jsonpath` to validate they are being passed from the mock service to the controller and out successfully. 
- You can replace `Arrays.asList` with newer `List.of` since Java 9.

```
    mockMvc.perform(get(url))
        .andExpect(status().isOk())
        .andExpect(jsonPath("$.[0].id").value(<>)
        .andExpect(jsonPath("$.[1].id").value(<>);
```

src/main/java/com/bravo/user/service/PaymentService.java
```diff
@@ -0,0 +1,42 @@
...
+    private final ResourceMapper resourceMapper;
+
+    public PaymentService(PaymentRepository paymentRepository, ResourceMapper resourceMapper) {
+        this.paymentRepository = paymentRepository;
+        this.resourceMapper = resourceMapper;
+    }
+
+    public List<PaymentDto> retrieveByUserId(
+            final String userId,
+            final PageRequest pageRequest,
```
tlynema-bravolt, line 32: With the values in pageRequest being able to be null.  Should this be optional?

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
```diff
@@ -0,0 +1,58 @@
...
+        when(paymentService.retrieveByUserId(anyString(), isNull(), isA(HttpServletResponse.class)))
+                .thenReturn(mockPaymentDtos);
+
+        String userId = "008a4215-0b1d-445e-b655-a964039cbb5a";
+        final ResultActions result = this.mockMvc
+                .perform(get("/payment/retrieve/" + userId))
+                .andExpect(status().isOk());
+
+        verify(paymentService).retrieveByUserId(eq(userId), any(PageRequest.class), any(HttpServletResponse.class));
```
wheelerswebservices, line 55: Why `any(PageRequest.class)`?

src/main/resources/data.sql
```diff
@@ -648,3 +648,20 @@ insert into address (id, user_id, line1, line2, city, state, zip) values
 ('42f33d30-f3f8-4743-a94e-4db11fdb747d', '008a4215-0b1d-445e-b655-a964039cbb5a', '412 Maple St', null, 'Dowagiac', 'Michigan', '49047'),
 ('579872ec-46f8-46b5-b809-d0724d965f0e', '00963d9b-f884-485e-9455-fcf30c6ac379', '237 Mountain Ter', 'Apt 10', 'Odenville', 'Alabama', '35120'),
 ('95a983d0-ba0e-4f30-afb6-667d4724b253', '00963d9b-f884-485e-9455-fcf30c6ac379', '107 Annettes Ct', null, 'Aydlett', 'North Carolina', '27916');
+
+INSERT INTO payment (id, user_id, card_number, expiry_month, expiry_year, updated) VALUES
+('0001', '008a4215-0b1d-445e-b655-a964039cbb5a', '4532016747591234', 1, 2024, NOW()),
+('0002', '008a4215-0b1d-445e-b655-a964039cbb5a', '6011514433546201', 2, 2025, NOW()),
+('0003', '008a4215-0b1d-445e-b655-a964039cbb5a', '6771549495588452', 3, 2023, NOW()),
```
tlynema-bravolt, line 655: Is this sample data for testing?  Not sure where this is used.

.gitignore
dschroeter26: Intellij Idea.

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
dschroeter26: Yes, there could be an empty 200, for when the user does not have any payment records.

src/main/java/com/bravo/user/controller/PaymentController.java
dschroeter26: No, but the instructions were to use a userId as a @PathVariable, I would prefer to use a JWT and pull a unique identifier from that and retrieve user data from some kind of data store.

src/main/java/com/bravo/user/dao/repository/PaymentRepository.java
dschroeter26: I created a custom method to add Pagination to the request, the last 3 years I have been doing Middleware BFF work and haven't used JPA as we are calling existing services and aggregating then for the client, so this would be where I have less experience.

I dont think it is, looks like I may have left that in when I was modeling this after the User resource.

src/main/java/com/bravo/user/service/PaymentService.java
dschroeter26: Ah yes it is, I was modeling this after the User Resource as per the instructions and forgot to change this.

src/main/resources/data.sql
dschroeter26: All valid, again I don't have the strongest SQL background as the majority of my experience is in service aggregation for a BFF architecture.

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
dschroeter26: All valid, I have been stuck on 1.8 at ESI since March 2020 lol

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
dschroeter26: Because I dont have the actual value, otherwise I would use that as a matcher, realistically I would use the real value on all matchers if available. But I may have taken the instructions to model this after the UserResource a little to _literally_, I figured _When In Rome_. This seems like a test between the controller test and the spring framework, so half integration and half unit test. I would decompose this into two separate tests, one integration where the spring context spins up all live beans and uses H2 and a Unit test where and injected dependencies are mocked.

src/main/resources/data.sql
dschroeter26: This is used by H2, an in memory database, you can access this data by running the app and sending a GET request to the Payments controller.

dschroeter26:
I think I took the instructions a little too literally, there are some things I would do different in the design of the REST api, it would be easier to talk about this on a call.

src/main/java/com/bravo/user/service/PaymentService.java
dschroeter26: I prefer to not have optional arguments, also I was modeling this after the UserResource and was following the same coding style and design that was already established. If I was designing this I would have to separate methods one that expects a PageRequest, and one that does not.

.gitignore
tlynema-bravolt: Thought that was just eclipse?  https://gist.github.com/fedir/cd58012d42c8c1edbb3fff611de6c2f6  

image

dschroeter26: I did open this in STS because my Intellij Ultimate trial ran out, then I downloaded Intellij CE and ran the app from the command line.

PR #48, opened 2023-08-15
Us add payment entity 2

jluisbn:


jluisbn:
Screenshot for existing user and existing payment
Retrieve_found_payment

Screenshot for existing user but no payment found
Retrieve_no_found_payment

Screenshot for user not found
Retrieve_user_no_found

src/main/java/com/bravo/user/controller/PaymentController.java
```diff
@@ -0,0 +1,41 @@
...
+@RestController
+@Tag(name = "Payment", description = "User payments")
+@RequestMapping(value = "/payment")
+@Slf4j
+public class PaymentController {
+
+    private final PaymentService paymentService;
+    private final UserValidator userValidator;
+
+    public PaymentController(PaymentService paymentService, UserValidator userValidator) {
```
wheeleruniverse, line 26: Do you know about `@RequiredArgsConstructor`?

src/main/java/com/bravo/user/controller/PaymentController.java
```diff
@@ -0,0 +1,41 @@
...
+        this.paymentService = paymentService;
+        this.userValidator = userValidator;
+    }
+
+    @Operation(summary = "Retrieves a list of payments by User")
+    @GetMapping("/retrieve/{userId}")
+    @ResponseBody
+    public List<PaymentDto> retrievePaymentsByUser(final @PathVariable String userId) {
+
+        log.info("Retrieving payments for user Id {}", userId);
```
wheeleruniverse, line 36: Do you think this `log` belongs after the validation?

src/main/java/com/bravo/user/dao/repository/PaymentRepository.java
```diff
@@ -0,0 +1,13 @@
...
+import com.bravo.user.dao.model.Payment;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+import java.util.List;
+
+@Repository
+public interface PaymentRepository extends JpaRepository<Payment, String> {
+
+    List<Payment> findAllByUserId(String userId);
```
wheeleruniverse, line 12: How is this different than `findByUserId`? Does `findByIgnoreCase` make sense here?

src/main/java/com/bravo/user/service/PaymentService.java
```diff
@@ -0,0 +1,37 @@
...
+    private final ResourceMapper mapper;
+
+    public PaymentService(PaymentRepository paymentRepository, UserRepository userRepository, ResourceMapper mapper) {
+        this.paymentRepository = paymentRepository;
+        this.userRepository = userRepository;
+        this.mapper = mapper;
+    }
+
+    public List<PaymentDto> retrieveByUserId(String userId) {
+        log.info("Retrieving all payments by user Id");
```
wheeleruniverse, line 29: It doesn't make sense to have a `log` at the Controller and Service layers. You are not passing the `userId` into the `log` message.

src/main/java/com/bravo/user/service/PaymentService.java
```diff
@@ -0,0 +1,37 @@
...
+
+        userRepository.findById(userId).orElseThrow(() -> new DataNotFoundException("User Id does not exist. Please validate input"));
```
wheeleruniverse, line 31: Does it make sense to declare a `existsById` query instead of doing a search to throw away the result. `exists` or `count` would be more efficient on the database.  Alternatively do we need to check at all? Is it fine to return 0 `payments` when the `userId` provided does not exist?

src/main/resources/data.sql
```diff
@@ -646,3 +646,11 @@ insert into address (id, user_id, line1, line2, city, state, zip) values
...
 ('95a983d0-ba0e-4f30-afb6-667d4724b253', '00963d9b-f884-485e-9455-fcf30c6ac379', '107 Annettes Ct', null, 'Aydlett', 'North Carolina', '27916');
+
+insert into payment (id, user_id, card_number, expiry_month, expiry_year, updated) values
+('15040055-a640-4cea-cfb7-357669fcdefe', 'fd6d21f6-f1c2-473d-8ed7-f3f9c7550cc9', '7219038435696266', 8, 2028, {ts '2023-07-01 07:20:16.267'}),
+('751c48b7-bbe4-4e94-bacb-31eb81fd2c12', 'f9096b14-daab-430f-835b-2f0d651e3559', '4390580235433205', 9, 2027, {ts '2022-02-11 09:55:48.371'}),
+('6033d04d-c602-4354-a411-d376a9215c68', 'f8bbbc7a-867b-4e8d-a2a5-8b7c07a4589c', '3997869182123667', 1, 2024, {ts '2020-11-12 11:22:15.264'}),
+('dd52bd06-d337-4624-d9bf-9593f2cff4c1', '06c70a5b-1950-4765-81b6-4bb2378e0991', '8439375960556557', 2, 2026, {ts '2018-05-11 17:22:15.125'}),
+('1628a7c9-ed7e-4582-eac6-b3321c2fdbe1', 'd54b6b5d-de10-4c06-9837-c7457937daf8', '5929456670566065', 3, 2025, {ts '2019-04-26 06:55:27.748'}),
+('6931d250-f972-4c5b-fc65-85904473de56', 'f8160123-1766-41c7-8056-2acbfd3c4f2a', '2981890323295572', 7, 2024, {ts '2022-08-08 09:11:52.428'});
```
wheeleruniverse, line 656: Can you explain this syntax? `{ts '2023-07-01 07:20:16.267'}` for the `updated` field?

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
```diff
@@ -0,0 +1,79 @@
...
+        final MvcResult result = this.mockMvc
+                .perform(get("/payment/retrieve/" + userId))
+                .andExpect(status().isOk())
+                .andReturn();
+
+        String responseBody = result.getResponse().getContentAsString();
+
+        assertNotNull(result);
+        assertNotNull(responseBody);
+        assertTrue(responseBody.contains("\"id\" : \"15040055-a640-4cea-cfb7-357669fcdefe\""));
```
wheeleruniverse, line 54: You can chain the `.getResponse().getContentAsString()` onto `result` since the `result` variable is only used that 1 time. 

```
final String responseBody = this.mockMvc
                .perform(get("/payment/retrieve/" + userId))
                .andExpect(status().isOk())
                .andReturn()
                .getResponse()
                .getContentAsString();
```

- Why is the `responseBody` variable not `final`? 
- The `assertNotNull(result)` is not necessary because the creation of `responseBody` would cause a NPE
- The `assertNotNull(responseBody)` is not necessary because the `assertTrue` would cause a NPE if the `responseBody` is `null` and the test will fail. 

Have you heard of `jsonpath`? https://www.petrikainulainen.net/programming/spring-framework/integration-testing-of-spring-mvc-applications-write-clean-assertions-with-jsonpath/

src/test/java/com/bravo/user/service/PaymentServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+@ContextConfiguration(classes = {App.class})
+@ExtendWith(SpringExtension.class)
+@SpringBootTest()
+class PaymentServiceTest {
+
+    private final PaymentRepository paymentRepository = mock(PaymentRepository.class);
+    private final UserRepository userRepository = mock(UserRepository.class);
+    private final ResourceMapper resourceMapper = mock(ResourceMapper.class);
+
+    private final PaymentService underTest = new PaymentService(
```
wheeleruniverse, line 34: Why are you using `mock` and `constructor` instead of `@MockBean`, `@Mock`, and `@InjectMocks`?

src/test/java/com/bravo/user/service/PaymentServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+            userRepository,
+            resourceMapper
+    );
+
+    @Test
+    void retrieveByUserId() {
+        final String userId = "user-id";
+
+        when(userRepository.findById(userId)).thenReturn(Optional.of(TestUtil.getUserData()));
+        when(paymentRepository.findAllByUserId(userId)).thenReturn(TestUtil.getPayments());
```
wheeleruniverse, line 45: Is this better than a looser matcher and a `verify`?  Do you know the difference? 

```
when(userRepository.findById(anyString())).thenReturn(Optional.of(TestUtil.getUserData()));

...

verify(userRepository).findById(userId);
```

src/test/java/com/bravo/user/service/PaymentServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+        final String userId = "user-id";
+
+        when(userRepository.findById(userId)).thenReturn(Optional.of(TestUtil.getUserData()));
+        when(paymentRepository.findAllByUserId(userId)).thenReturn(Collections.emptyList());
+        when(resourceMapper.convertPayments(Collections.emptyList())).thenReturn(Collections.emptyList());
+
+        List<PaymentDto> payments = underTest.retrieveByUserId(userId);
+
+        assertNotNull(payments);
+        assertEquals(payments.size(), 0);
```
wheeleruniverse, line 65: What about `assertEmpty`?

src/test/java/com/bravo/user/util/TestUtil.java
```diff
@@ -0,0 +1,49 @@
...
+    public static User getUserData() {
+
+        UserSaveDto userDto = new UserSaveDto();
+
+        userDto.setFirstName("Harrison");
+        userDto.setMiddleName("");
+        userDto.setLastName("Ford");
+        userDto.setPhoneNumber("");
+
+        return new User(userDto);
```
wheeleruniverse, line 22: Why did you create `UserSaveDto` instead of `User` directly? These variables can be `final` and if you use `@Builder` you can remove the variable altogether.

wheeleruniverse:
Thank you so much! I left some comments on the PR for your benefit. We don't expect any additional commits so don't feel obligated to do so. Let me know if you have any questions or comments. Please respond to the questions I left in your code if you wouldn't mind. I would love to get your thoughts.

src/main/java/com/bravo/user/controller/PaymentController.java
jluisbn: I do, I just forgot, and also tried to follow UserController convention as requested.

src/main/java/com/bravo/user/controller/PaymentController.java
jluisbn: For sure, I just think I like to keep trail of everithing that happens, so it is easier to debug in my opinion.

src/main/java/com/bravo/user/dao/repository/PaymentRepository.java
jluisbn: I just use that more descriptive even though it is the same result as byUserId. And about the ignoreCase, I think we should look for exact keys, and also we could face some performance issues if we start doing searches with ignore case, let's say we have an index in the userId, as far as I am concerned this type of searches would ignore the index and probably be an slow query.

src/main/java/com/bravo/user/service/PaymentService.java
jluisbn: Agreed, I just think I like to keep trail of everithing that happens, so it is easier to debug in my opinion.

src/main/java/com/bravo/user/service/PaymentService.java
jluisbn: Yes, you are right, maybe I am to defensive so if something does not look right I rather throw an exception, and let the user give it another try.

src/main/resources/data.sql
jluisbn: I thought it was just a timestamp as when the user updated their information. Probably should I paid more attention to it.

src/test/java/com/bravo/user/controller/PaymentControllerTest.java
jluisbn: You are right. Will keep that in mind.

src/test/java/com/bravo/user/service/PaymentServiceTest.java
jluisbn: This is just the way I have been doing this. But I think using annotations is a cleaner way.

src/test/java/com/bravo/user/service/PaymentServiceTest.java
jluisbn: In my opinion that depends, because if you have a more complex method then you have to mock every call or access to their properties, but if you send a complete object (not a mock) then your code will not throw a NPE, and instead of having many "when then" I think it is a bit cleaner to have real objects.
And about the verify, I forgot to add that, I tend to always add it.

src/test/java/com/bravo/user/service/PaymentServiceTest.java
jluisbn: I have not used that one, and since I tend to use Sonarlint to avoid this code smells, Intellij did not suggest that method.

src/test/java/com/bravo/user/util/TestUtil.java
jluisbn: Yes, I missed that one.

